PR #5, opened 2018-09-04
systemd timers

wyattjoh:
systemd timers need to be started as well as enabled, otherwise the timer won't actually be registered via `systemctl is-active restic-backup.timer`.


erikw:
You are right. I forgot to add this, thanks!

PR #6, opened 2018-09-24
<README: Add some missing word>

bobsaintcool:
Hey! 

I think that some words are missing from the docs, not sure but I think I'm close of what you would like to say.
---
Give a big overview on how proposed automated backup works.


erikw:
Yes, thanks!

Issue #12, opened 2018-10-27
Easier configuration


lahdekorpi:
Would it be ok to move the configuration related environmental variables from the `/usr/local/sbin/restic_backup.sh` to something like `/etc/restic/backup_env.sh` to make it more "unix like" and somewhat easier to configure?  
Clearly a very minor change, but just wanted to check before making a PR.

These:
```bash
RETENTION_DAYS
RETENTION_WEEKS
RETENTION_MONTHS
RETENTION_YEARS
BACKUP_PATHS
BACKUP_EXCLUDES
```

erikw:
Good idea! I would pick a new file rather than `/etc/restic/backup_env.sh` though, as this one is also used on interactive command line to work with `restic(1)` by sourcing this file.

Maybe a new file `/etc/restic/bacup_settings.sh` or the like, that is sourced from the backup scripts?

peperunas:
Hello!
I have been using "env" files to manage multiple backups, [click](https://github.com/peperunas/configs/tree/master/restic). My method uses systemd's "@" feature to load different configs for restic. 
Maybe we could move the configuration to external files as @lahdekorpi suggested and then use systemd's "@" operator.

Does it make sense to you?

lahdekorpi:
@peperunas Sound great!

Maybe we just load the configuration (or "env") file from the restic directory under etc with the name provided after the systemd @ character?  
Should be simple enough to implement.

erikw:
Loading different config sound like a good idea 👍

peperunas:
@lahdekorpi that was exactly what I was thinking. :-)

lahdekorpi:
I'm now thinking if making even the repository type for restic configurable. Or at least easy to extend.  
But a separate `restic_backup.sh` file for each type or a lot of if statements..? 🤔

erikw:
@lahdekorpi Hm at this still small size, I think we could stick to one script still, but utilize shell functions for code structure, check out out my repo here for example: https://github.com/erikw/tmux-powerline/blob/master/segments/now_playing.sh#L59-L72

mfeif:
This is sounding great! I have 3 different B2 repositories, each mapped to a mountpoint, so this software AS-IS is hard for me to use, but if I could have 3 different env sets, that would be excellent.

Can I help push this along? My bash foo is not great... but I can try.

erikw:
As a note to self; after merging this, the AUR PKGBUILD must be updated with the new file paths:

[...]

PR #14, opened 2019-01-07
Multi repos, pull env variables out of scripts

mfeif:
I've made a quick fork to pull env vars out of the scripts, and to move some other configuration into env files, as per the suggestion in #12.

Now, in /etc/restic there is a `_global.env` for things that are used across restic tasks, and `example.env` that should be renamed to a logical local name. Whatever that name is needs to correspond to an entry in /etc/systemd/system:

I use symlinks with the @ hack also mentioned by @peperunas in #12 so to make example function you'd make a link from restic-backup@.service to restic-backup@example.service (and the corresponding timers, check, etc.).

Moving the particulars into .env files allows us to put the paths, retention, and all of that stuff into a path-by-path or repo-by-repo config.

I'm not great at systemd or bash, so I'd love it if others could vet this. But it is working on my Ubuntu 18.10 box.

Thanks for the work, @erikw!


etc/restic/_global.env
```diff
@@ -0,0 +1,16 @@
+# B2 credentials.
+# Extracted settings so both systemd timers and user can just source this when want to work on my B2 backup.
+# See https://restic.readthedocs.io/en/latest/030_preparing_a_new_repo.html
+
+# these variables are sourced FIRST, and any values inside of *.env files for
```
erikw: It would be nicer maybe if instead the `*.env` files sources this global, think "import base class". Then the env files would be self-contained in the sense that you can do

```bash
$ . /etc/restic/example.env
$ restic snapshots
```

instead of having to source 2 files:

```bash
$  . /etc/restic/_global.env
$ . /etc/restic/example.env
$ restic snapshots
```

Being able to do this is very handy when you set up the repo, or want to work with it manually.

etc/restic/_global.env
```diff
@@ -0,0 +1,16 @@
...
+# specific configurations will over-ride if also defined there
```
erikw: ... thus add here as first line

```
source /etc/restic/_global.env
```

etc/systemd/system/restic-backup@.service
```diff
@@ -6,5 +6,5 @@ OnFailure=status-email-user@%n.service
 Type=simple
 Nice=10
 ExecStart=/usr/local/sbin/restic_backup.sh
-# $HOME or $XDG_CACHE_HOME must be set for restic to find /root/.cache/restic/
-Environment="HOME=/root"
\ No newline at end of file
+EnvironmentFile=-/etc/restic/_global.env
```
erikw: What does the `-` mean here?

etc/restic/_global.env
mfeif: I thought about that, but what stopped me was that I thought they'd need different syntax; one would have the "export FOO" and the .env would just have the declarations with the export.

I'm not attached to either way, though.

But why would it be "nicer"?

(By the way, they are now self-contained; the minus sign before the path in the .service file means that if it's not present it won't error.)

etc/restic/_global.env
mfeif: don't need to; the .system file grabs it:

```
Description=Backup with restic to Backblaze B2
OnFailure=status-email-user@%n.service

[Service]
Type=simple
Nice=10
ExecStart=/usr/local/sbin/restic_backup.sh
EnvironmentFile=-/etc/restic/_global.env
EnvironmentFile=/etc/restic/%I.env
```

etc/systemd/system/restic-backup@.service
mfeif: It means that if the file isn't there, it won't error (I looked it up too)

etc/restic/_global.env
erikw: I'm not sure what you mean with different syntax? 

_global.env:
```
B2_ACCOUNT_ID="<b2-account-id>"
B2_ACCOUNT_KEY="<b2-account-key>"
....
```

example.env:
```
source /etc/restic/_global.env

RESTIC_REPOSITORY="b2:<b2-repo-name>"
...
```

and now you only have to source `example.env` to get all settings, and the base vars are in the root `_global.env` so that they can be shared across other *.env files


Its' nicer because you only have to source 1 file to get all vars you need. See example in my comment :)

etc/restic/_global.env
erikw: I know. The intention is to make it self-contained:

```bash
$ . /etc/restic/example.env
$ restic snapshots
```

vs

```bash
$ . /etc/restic/_global.env
$ . /etc/restic/example.env
$ restic snapshots
```

etc/systemd/system/restic-backup@.service
erikw: I guess it should error, as it's needed? Either way I still think it's nicer to not spread the knowledge of the global around, and contain the configuration from one access point. Makes this file smaller too :)

etc/restic/_global.env
```diff
@@ -0,0 +1,16 @@
...
+
+# $HOME or $XDG_CACHE_HOME must be set for restic to find /root/.cache/restic/
+HOME=/root
+
+RESTIC_PASSWORD_FILE="/etc/restic/pw.txt"
+B2_ACCOUNT_ID="<b2-account-id>"
+B2_ACCOUNT_KEY="<b2-account-key>"
+
+# How many network connections to set up to B2. Default is 5.
+B2_CONNECTIONS=10
```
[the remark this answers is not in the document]
mfeif: I don't know what a sane number is here; the other branch from the pull request was 50! I just have a normal home internet service, so I turned it down.

But yeah, it is in env now.

etc/restic/_global.env
mfeif: I just re-read it; I see why it's nicer: manual. 

What do you think about the export thing; am I wrong about that?

mfeif: your old files had "export" for the variables; would that be necessary if we sourced it?

erikw: I'm not sure I understood how you meant. Like this?:

_global.env
```bash
export RESTIC_REPOSITORY
```

example.env:
```bash
RESTIC_REPOSITORY="the value"
```

etc/restic/_global.env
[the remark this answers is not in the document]
erikw: I guess 10 is fine :)

etc/systemd/system/restic-backup@.service
mfeif: Agreed.

When I was reading about systemd, it may have said that the EnvironmentFile isn't a shell script; it's just variables; it may not be able to run the source command. I'll have to try it.

mfeif: Also, it's not actually needed, necessarily: if all config is specific to the .env file, you don't need globals.

But I agree that making this compatible with manual scripting should be a goal. I need to try out some stuff and see.

etc/restic/_global.env
erikw: Aha, if `export` is needed. Possibly not, but I was thinking it does not hurt. You never know when you will drop to a subshell, or put a command in a subshell. Then it's nice to retain the variables :)

erikw:
@mfeif I see, thanks for checking.

This is currently a feature that that exist, that you can easily source the environment to your local shell, because even if the backups are automated, you are always going to want to do manual inspection to look at snapshots, or most prominently restore from backups. That's why I want to make sure we keep this feature.

I quickly made something that seems to work. We could include this file in the repo, and write in the repo that you should source it in your shell's startup script, if you want an easy way to load the environment to you shell


`load_restic_env.sh`:
```bash
# Source this file from you shell's statup script!
#
# After doing so, you get access to the function `restic_env` that will load the envionment for you, so you can work with the restic command wintout having to set all arguments like --repo etc.
#
# Example of usage:
# $ restic_env my_backup
# $ restic snapshots

# Source and export variables found in an env file.
__source_env_file() {
	envfile="$1"
	while read line; do
		if [[ $line =~ ^\s*.*=.*\s*$ ]]; then
			eval "export $line"
		fi
	done < $envfile
}

# Source restic envionment varialbes to current shell for a given envionment (/etc/restic/*.env)
restic_env() {
	local env_name="$1"
	__source_env_file /etc/restic/_global.env
	__source_env_file /etc/restic/${env_name}.env
}
```

mfeif:
This looks fantastic! I'll add it to my PR, adjust the README, and re-submit.

Seems like putting it into /usr/local/sbin with the other scripts makes sense.

Thanks!

mfeif:
Hi Erik.

I was getting back to this, it looks very promising.

I've hit a snag that I don't immediately know how to solve involving sudo. The env functions you suggest definitely work (I'm sourcing it via my ~/.profile). But those variables don't make it down to sudo:

```
$ restic_env manual
$ echo $RESTIC_REPOSITORY
b2:dude-backups-manual
$ restic snapshots
Resolving password failed: Readfile: open /etc/restic/restic-pw.txt: permission denied
$
$ sudo restic snapshots
Fatal: Please specify repository location (-r)
$ sudo restic_env manual
sudo: restic_env: command not found
```

This makes sense, in a way. So I tried putting the source commands into .profile in /root/ but apparently a sudo command doesn't run a shell; sudo restic just runs restic. I'm sure that there is a combination of switches to sudo that will pass through environment, but it seems to be a nasty UI to ask people to use sudo in a new way.

I've tried `sudo -s` but that clears out the environment too. One can run an interactive shell with `sudo bash` but that doesn't get us to `sudo restic {blah}`. 

I know that one can just get into a shell and `source /usr/local/sbin/load_restic_env.sh` or put that someplace that will get executed by a login shell, but... well, you see.

Any pointers before I go down a huge rabbit hole on this? Thanks

erikw:
@mfeif Ah.. I should have been clearer; I was assuming that this would be put in to the root user's shell profile, e.g. `/root/.profile`, and that you then would work with restic as root.

```bash
$ sudo -s
root$ restic_env my_backup
root$ restic snapshots
```

While I think it's possible, it seems like a bit of a hazzle to get it to work with `sudo`, so I would say let's keep it simple and ship this version that works for root shell, and improve on it later.

mfeif:
Yeah, I figured, but `sudo -s` doesn't source /root/.profile; I guess it's technically not an 'interactive' shell. `sudo -i` is, but it seems that most people don't use that or know about that; I sure didn't.

https://www.sudo.ws/man/1.8.17/sudo.man.html#i

One could argue that getting this worked out is out of scope, and an exercise for the reader on how to integrate shells, but I'd like to get workable instructions in there for people.

What do you think about just suggesting -i and linking to the docs?

mfeif:
I'm coming back to this, working on the README file; I see that you have a makefile and it's in the setup docs. I didn't look into modifying the makefile, and I'm not expert in that domain. Is this something you'd like to amend, or just remove and describe narratively how to install?

I'll keep working on the README meanwhile.

erikw:
It should be an easy matter to update. But if you don't want to I can update the Makefile after your PR is merged :)

jamesfraser:
Hi all,

If you want to get the environment variables passed through to the script, there is the [PassEnvironment=](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#PassEnvironment=) option.

It's as simple as adding lines to the .service unit files as follows (it could all be done in one line, but I find splitting them up is more readable):
```
 PassEnvironment=HOME RESTIC_PASSWORD_FILE B2_ACCOUNT_ID B2_ACCOUNT_KEY
 PassEnvironment=RESTIC_REPOSITORY RETENTION_DAYS RETENTION_WEEKS RETENTION_MONTHS RETENTION_YEARS
 PassEnvironment=BACKUP_PATHS BACKUP_EXCLUDES BACKUP_TAG
```

Hope this helps!

@erikw thanks for your time in releasing these!

gerardbosch:
> @mfeif Ah.. I should have been clearer; I was assuming that this would be put in to the root user's shell profile, e.g. `/root/.profile`, and that you then would work with restic as root.
> 
> ```shell
> $ sudo -s
> root$ restic_env my_backup
> root$ restic snapshots
> ```
> 
> While I think it's possible, it seems like a bit of a hazzle to get it to work with `sudo`, so I would say let's keep it simple and ship this version that works for root shell, and improve on it later.

Hi there! I'm not sure if the following could be useful, but for my own convenience I've been using the following for some time to run restic and to load environment variables sourced from `/etc/restic/b2_env.sh` (which is owned by root):

```bash
❯ cat ~/bin/resticw                 
#!/bin/bash

#
# A little restic wrapper just to load restic environment vars
# from /etc/restic in a subshell so they are not exposed
#

if [ $(id -u) -ne 0 ]; then
  echo "Elevating to root to load restic environment..."
  exec sudo bash "$0" "$@"
fi

# Source environment in a subshell and run restic
( source /etc/restic/b2_env.sh && restic "$@" )
```

`resticw` stands for restic wrapper.
This way I don't need to care on how sourcing environment, I just do:

```shell
$ resticw snapshots
```

My understanding was (may I'm wrong) that in this way credentials are never exposed to the shell beyond the temporary subshell run by sudo.

Do you think something like that would make sense or help? Maybe a more sophisticated wrapper to handle different profiles/environments as this PR is adding? e.g. `resticw profileA snapshots`. Just an idea!

mfeif:
Hi! I let this drop, but I'm still using it and it's working... I'll put it back on my to do list!

gerardbosch:
Hi @erikw, @mfeif, I was interested in improving the configuration and segregate it from the backup script. As it's being a while since last update, I continued from this PR, solving conflicts and expanding stuff. I'm publishing it here: https://github.com/erikw/restic-systemd-automatic-backup/pull/45

erikw:
I'm closing this as the work was continued based on this branch in #45

PR #15, opened 2019-04-26
Use file templates to improve security, reproducibility

billwanjohi:
By git-ignoring the files that will hold passwords, creating associated template files for them, and then conditionally creating the files with a couple of new make targets, we can help people install their configuration in a reproducible manner without compromising on security. The commits are separated logically, so you can cherry-pick what you see fit.


Makefile
```diff
@@ -30,12 +30,19 @@ install: install-scripts install-conf install-exclude install-systemd
...
+
+etc/restic/b2_pw.txt:
+	install -m 0600 etc/restic/b2_pw.txt.template etc/restic/b2_pw.txt
 
 # target: install-conf - Install restic configuration files.
-install-conf:
-	install -d $(DEST_CONF) -m 700
-	install $(SRCS_CONF) $(DEST_CONF)
+# will create these files locally only if they don't already exist
+install-conf: | etc/restic/b2_env.sh etc/restic/b2_pw.txt
```
billwanjohi: the pipe delimiter indicates that the succeeding files are [order-only prerequisites](https://stackoverflow.com/a/21754930/684619), meaning they will only be made if they don't already exist

Makefile
```diff
@@ -44,4 +51,4 @@ install-exclude:
 # target: install-systemd - Install systemd timer and service files
 install-systemd:
 	install -d $(DEST_SYSTEMD)
-	install -m 0644 $(SRCS_SYSTEMD) $(DEST_SYSTEMD)
+	install $(SRCS_SYSTEMD) $(DEST_SYSTEMD)
```
billwanjohi: I found this file permission to be redundant, since this is what I get by default on new files placed in this destination.

erikw:
Nice contribution!

Issue #16, opened 2019-07-12
Check for metered network


toddejohnson:
On my laptop if I'm tethered or on a limited connection the backup should not run. NetworkManager has a way to check for this. My solution is based upon https://github.com/jdorel/systemd-metered-connection-dependency as I use systemd. 

Would you accept a pull request based upon the above?

erikw:
Discussion continues in PR :)

PR #17, opened 2019-07-13
Add requires to unmetered-connection.service

toddejohnson:
This implements #16 and adds a dependency to need https://github.com/jdorel/systemd-metered-connection-dependency installed.


erikw:
I would not like to add a mandatory dependency by default. But we could make a copy of the service and call it `-metetchecked` or something, as a drop in for those that prefer this.

Either way README needs instructions on how to install the dependency. Even better if it's done automatically. Make a option to make can chose to set up the metered version

toddejohnson:
I totally understand. I'm sure @jdorel wouldn't have an issue with us using his idea in this. I'll work on this more tonight...

etc/systemd/system/restic-backup.timer
```diff
@@ -1,5 +1,6 @@
 [Unit]
 Description=Backup with restic on schedule
+Requires=nm-unmetered-connection.service
```
toddejohnson: I'm not sure if this is actually needed but doesn't seem to hurt.

erikw:
Okay better, no install dependency :). Now what if you don't use network manager though then the check would always fail or?? I don't use NetworkManager. 

I would like unmeteted NetworManager to be opt-in,  maybe with an option to Make that installs the service with or without the unmeteted part. Or at least make the unmeteted script return 0 if no NetworkManager is used on the system

toddejohnson:
My thoughts exactly. I'm going to write some checks for dbus and NetworkManager into the script and return success if they aren't in use. 

I thought about making it optional and am open to suggestions. Less options more just works out of the box is my style. Sleep got me before I got the checks added.

toddejohnson:
I'm sorry for the delay. I'm trying to pick how to best implement this for the systemd platforms I understand (Arch Linux, EL7, Some debian 9)

I'm leaning towards the below as I'm already attached to systemd:
```sh
systemctl is-active dbus.service >/dev/null 2>&1 || exit 0
systemctl is-active NetworkManager.service >/dev/null 2>&1 || exit 0
```
An alternate option that might be less systemd attached is below however this is all systemd based for now:
```sh
[ $(ps -e | grep dbus | grep -v grep | wc -l) -gt 0 ] || exit 0
[ $(ps -e | grep NetworkManager | grep -v grep | wc -l) -gt 0 ] || exit 0
```
What do you think @erikw?

erikw:
@toddejohnson The first option looks good to me :-)

PR #18, opened 2019-07-13
Fix missing directories on install

toddejohnson:
On my system without changes I'd get errors with missing /mnt/media and /home/erikw/.backup_exclude folders/files.  To further simplify installs and remove the need to edit restic_backup.sh lets add some checks to non standard BACKUP_PATHS and  BACKUP_EXCLUDES. This will help till #12 and #14 can be worked out.


usr/local/sbin/restic_backup.sh
```diff
@@ -22,8 +22,17 @@ RETENTION_MONTHS=18
 RETENTION_YEARS=3
 
 # What to backup, and what to not
-BACKUP_PATHS="/ /boot /home /mnt/media"
-BACKUP_EXCLUDES="--exclude-file /.backup_exclude --exclude-file /mnt/media/.backup_exclude --exclude-file /home/erikw/.backup_exclude"
+BACKUP_PATHS="/ /boot /home"
+[ -d /mnt/media ] && BACKUP_PATHS+=" /mnt/media"
+BACKUP_EXCLUDES="--exclude-file /.backup_exclude"
```
erikw: This needs to be changed to /etc/restic/ since the other PR

toddejohnson: Good catch! I was nervous which PR you would accept so I made them all individual. I'll have this fixed in a moment...

erikw:
Looks better!

PR #20, opened 2019-07-16
Set permissions for systemd units to 644

toddejohnson:
Systemd Units should not be executable:
```
Jul 13 23:52:41 systemd[1]: Configuration file /usr/lib/systemd/system/restic-check.service is marked executable. Please remove executable permission bits. Proceeding anyway.
Jul 13 23:52:41 systemd[1]: Configuration file /usr/lib/systemd/system/restic-check.timer is marked executable. Please remove executable permission bits. Proceeding anyway.
```
On Arch they are 0755 so we need to force 0644. This is a bit of a reversion of #15 where we assumed it would be in /etc/ and the default would be acceptable.


erikw:
Thanks!

PR #22, opened 2019-07-26
Forget now has --prune which might be more efficent.

toddejohnson:
I was comparing one of my servers and it uses restic forget --prune instead of restic prune. It appears a bit more efficient as it [only runs prune if there is something to prune. ](https://restic.readthedocs.io/en/stable/060_forget.html?highlight=--prune)

### DIFF
--- a/usr/local/sbin/restic_backup.sh
+++ b/usr/local/sbin/restic_backup.sh
@@ -71,7 +71,8 @@ wait $!
 restic forget \
 	--verbose \
 	--tag $BACKUP_TAG \
-        --prune
+	--option b2.connections=$B2_CONNECTIONS \
+        --prune \
 	--group-by "paths,tags" \
 	--keep-daily $RETENTION_DAYS \
 	--keep-weekly $RETENTION_WEEKS \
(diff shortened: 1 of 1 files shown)


usr/local/sbin/restic_backup.sh
```diff
@@ -71,20 +71,14 @@ wait $!
...
 	--keep-daily $RETENTION_DAYS \
 	--keep-weekly $RETENTION_WEEKS \
 	--keep-monthly $RETENTION_MONTHS \
 	--keep-yearly $RETENTION_YEARS &
 wait $!
 
-# Remove old data not linked anymore.
-# See restic-prune(1) or http://restic.readthedocs.io/en/latest/060_forget.html
-restic prune \
-	--option b2.connections=$B2_CONNECTIONS \
```
erikw: This should maybe be migrated to restic-forget?

toddejohnson: I've added the B2 connection limit to restic forget in 429ee40.

toddejohnson:
I've updated the comment and the forget.

PR #26, opened 2020-01-22
README: Fix backup_exclude destination

shwang:
restic_backup.sh doesn't actually look for `/.backup_exclude` it seems


erikw:
You're right. Thanks!

PR #27, opened 2020-06-02
Fix conflicts for restic-check service

wvolz:
This fixes a typo in the restic-check.service file that makes sure a backup is not running before starting a check.


erikw:
Great, thanks for the fix!

PR #37, opened 2021-10-29
Makefile: add uninstall target

erikw:
Fixes #30

PR #39, opened 2021-10-30
Update README: Add uninstall instructions

gerardbosch:
Just update the README with the uninstall addition from https://github.com/erikw/restic-systemd-automatic-backup/pull/37.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,3 +128,11 @@ If you want to run an all-classic cron job instead, do like this:
 
 * `etc/cron.d/restic`: Depending on your system's cron, put this in `/etc/cron.d/` or similar, or copy the contents to $(sudo crontab -e). The format of this file is tested under FreeBSD, and might need adaptions depending on your cron.
 * `usr/local/sbin/cron_mail`: A wrapper for running cron jobs, that sends output of the job as an email using the mail(1) command.
+
+## Uninstall
+
+There is a make target to remove all files (scripts and configs) that were installed by `sudo make install`. Just run:
+
+```console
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -128,3 +128,11 @@ If you want to run an all-classic cron job instead, do like this:
 
 * `etc/cron.d/restic`: Depending on your system's cron, put this in `/etc/cron.d/` or similar, or copy the contents to $(sudo crontab -e). The format of this file is tested under FreeBSD, and might need adaptions depending on your cron.
 * `usr/local/sbin/cron_mail`: A wrapper for running cron jobs, that sends output of the job as an email using the mail(1) command.
+
+## Uninstall
+
+There is a make target to remove all files (scripts and configs) that were installed by `sudo make install`. Just run:
+
+```shell
```
erikw, line 136: Just minor, to be consistent with the rest of the readme, use
````
```console
$ sudo make uninstall
```
````

I somewhat recently learned that probably `console` should be used for command line segments, rather than `bash` or `shell`.

:)

erikw (changes requested):
Thanks!

Just a minor edit requested :)

README.md
gerardbosch: Sure! no problem, I was just used to use `shell` and wrote it by habit. I also added the prompt `$`, I didn't do it at first as sometimes is easier to copy/paste without prompt, and nowadays more people uses fancy prompts which are not the default bash `$`.

I've just updated and force pushed the PR.

PR #45, opened 2022-01-16
Multiple profile configuration

### LINKED ISSUE #12: Easier configuration

gerardbosch:
Improve organisation of configurations by extracting them to different environment files that enable different "backup profiles". Profiles specify a set of configurations and are loaded by passing them as a systemd parameter.

Resolves #12


gerardbosch:
Hi @erikw, if we move ahead with this PR I would suggest to go with "squash-merge" as the log became a bit convoluted. Doing this the PR will be still co-authored I think :) Cheers!

usr/local/sbin/restic_backup.sh
```diff
@@ -18,9 +18,9 @@ trap exit_hook INT TERM
 
 # Set up exclude files: global + path-specific ones.
 exclusion_args="--exclude-file /etc/restic/backup_exclude"
-for homedir in /home/*; do
-	if [ -f "$homedir/.backup_exclude" ]; then
-		exclusion_args+=" --exclude-file $homedir/.backup_exclude"
+for backup_path in ${BACKUP_PATHS[@]}; do
```
gerardbosch, line 25: Question here: What will happen if `BACKUP_PATHS=/home/` and, then there is a `/home/someuser/.backup_exclude` file? The file makes reference to a specific user, while the paths refer to the higher level directory `/home`.

The previous behaviour takes the `someuser` exclusion file in consideration, but after the change what will happen? I think it will be omitted? If that's right, may be confusing. Did I get it right? What do you think? :D

Maybe aside of `backup_paths/.backup_exclude`, also all user homedir `.backup_exclude` can be considered.

erikw: Ah this was the original intention, got-cha! Thanks

I had forgotten. We should look for `.backup_exclude` for each mount-point, and additionally each user dir in `/home`, if `/home` is in the backup path.

Will fix!

gerardbosch: Cool 😊

mfeif:
Thanks @gerardbosch for picking this up and running with it. I'm (also) having trouble navigating all of the changes since 2019, so I'll start from scratch and install this proposed merge on a new system and see if I find anything confusing or broken.

Good stuff! Thanks!

erikw:
@mfeif Hallo!

I'm happy to hear that you want to try this out. Perfect!

Could you maybe wait just 1 day or so, as there are still some changes coming in to this PR :)

But soon after that, having someone trying out the updated setup from scratch would be terrific!!

mfeif:
Of course!

usr/local/sbin/restic_backup.sh
erikw: So, please check 424bb0c117285de5a3a22aee35f37544ec4a6fcf

I tried it out locally, seemed to work... 😅

erikw:
# Review status from me

- [x] Review changes
- [x] Contribute to README & small fixes
- [ ] Test using the updated Makefile
- [ ] Test using the new profiles on an actual system
- [ ] After merge: Update [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=restic-systemd-automatic-backup)
- [ ] After merge: make new major release tag

@gerardbosch as I reviewed the changes I asked myself if it would make sense to make the `/etc/restic/pw.txt` & `/etc/restic/backup_exclude` not global but separate per profile. Maybe one would like to use different passwords for each profile? This is not a use case for me, as I'm using one profile only, but the question came up for me. Maybe this just makes it more complicated though

mfeif:
> @gerardbosch @mfeif as I reviewed the changes I asked myself if it would make sense to make the `/etc/restic/pw.txt` & `/etc/restic/backup_exclude` not global but separate per profile. Maybe one would like to use different passwords for each profile? This is not a use case for me, as I'm using one profile only, but the question came up for me. Maybe this just makes it more complicated though

That seems an obvious thing that someone will eventually need/want. Good thinking. Putting those in .env files might be messy, so separate text files might be nice.

Can we still cascade well, in the event that there is NOT a per-target customization? Without the natural over-ride mechanism with ENV variables?

erikw:
I think it should work with the same approach for overriding `_default.env` in a `profile.env`. To make it more clear, I'd suggest that we:

* Rename `/etc/restic/pw.txt` to `/etc/restic/pw_global.txt`
* Rename `/etc/restic/backup_exclude` to `/etc/restic/backup_exclude_global.txt`
* Create a variable `RESTIC_EXCLUDE_FILE=/etc/restic/backup_exclude_global.txt` and put it in `/etc/restic/_global.env` and use it in `restic_backup.sh`.
* Add two out-commented lines in the `default.env` just to demonstrate that this feature exist if needed
   ```bash
   # In simple setups e.g. only one backup profile, having a single password and exclude file is fine.
   # If you have multiple profiles and desire to have separate password for each, or exclude files, you can override _default.env values like this:
  #export RESTIC_PASSWORD_FILE="/etc/restic/pw_myprofile.txt"
  #export RESTIC_EXCLUDE_FILE=/etc/restic/backup_exclude_myprofile
   ```

gerardbosch:
> @gerardbosch @mfeif as I reviewed the changes I asked myself if it would make sense to make the /etc/restic/pw.txt & /etc/restic/backup_exclude not global but separate per profile. Maybe one would like to use different passwords for each profile? This is not a use case for me, as I'm using one profile only, but the question came up for me. Maybe this just makes it more complicated though

Hi there :)

If I'm not wrong, this (at least the password per profile part) can be achieved out-of-the-box, couldn't it? I mean, any profile `foo-profile.env` is sourcing the `_global.env`, so for that particular use case, just re-declaring `RESTIC_PASSWORD_FILE=/path/to/foo-profile_pw.txt` @ `foo-profile.env` will override the "global" value –that's even stated in the `default.env` comments.

---
As per the "per profile exclusion file" it should work the same, but in this case, this looks hardcoded into `restic_backup.sh`: 

```bash
##  Global backup configuration.
exclusion_args="--exclude-file /etc/restic/backup_exclude"
```

so I think it would be good to put that path on a variable too (`RESTIC_EXCLUDE_FILE`), I can do it if you want :)

---

My opinion (in case it helps :D) is to focus the structure and file names to the most common use case, that's backup a single computer or account. Then, as the facilities are there, just re-declaring a couple of env vars gives flexibility beyond that.

Even I've just thought, you could use several profiles with different mail/`B2_ACCOUNT_ID` and get a multiple 10GB free storage per profile ^_^

erikw:
@gerardbosch Thanks for the input. Let's go with just adding the `RESTIC_EXCLUDE_FILE` and use it in `restic_backup.sh` as well. Go ahead :)

usr/local/sbin/restic_backup.sh
```diff
@@ -15,39 +20,36 @@ exit_hook() {
...
+			excl_xdg="$homedir/.config/restic/backup_exclude"
+			test -f $excl_home && args+=" --exclude-file $excl_home"
+			test -f $excl_xdg && args+=" --exclude-file $excl_xdg"
+		done
+	fi
+	echo $args
+}
+exclusion_args="$exclusion_args $(backup_files_from_homedirs /home)"
+## And the same service of macOS users, having /Users instead of /home
+exclusion_args="$exclusion_args $(backup_files_from_homedirs /Users)"
```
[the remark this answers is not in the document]
erikw: Accidentally it will have a leading space if it returns anything, else empty string, so it would work, if this contract stays true. I guess this is fine, both works :)

usr/local/sbin/restic_backup.sh
```diff
@@ -15,39 +20,36 @@ exit_hook() {
...
-
-
-# Set all environment variables like
-# B2_ACCOUNT_ID, B2_ACCOUNT_KEY, RESTIC_REPOSITORY etc.
-source /etc/restic/b2_env.sh
-
-# How many network connections to set up to B2. Default is 5.
-B2_CONNECTIONS=50
+## Additonal service: if /home is being backed up, allow user specifci backup files in home directory or default $XDG_CONFIG_HOME path
+backup_files_from_homedirs() {
```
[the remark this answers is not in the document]
erikw: Makes a lot of sense, much better name!

usr/local/sbin/restic_backup.sh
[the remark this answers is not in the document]
erikw: @gerardbosch out of interest: how did you make these changes appear as "Approve these suggestion" button for me?

I don't think I've seen this feature in GitHub before. Did you make the code suggestion within the GitHub UI, or is there a way to push a commit as a "suggestion" to a PR? :)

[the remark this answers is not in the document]
gerardbosch: Hey @erikw There's a button in the GitHub website to insert suggestions. It works very nice when you perform a comment over a bunch of lines (you can click the `+` button and hold, then drag down when initiating a comment).

The sometimes "danger of this" is that it is not an IDE, and sometimes, accepting a suggestion, leaves stuff not properly updated behind, so use with caution :)

image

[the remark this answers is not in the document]
gerardbosch: :point_up: that's exactly what happened here, hehe https://github.com/erikw/restic-systemd-automatic-backup/pull/45/commits/782afa9b0a0315729d3b9a64e818d7a064fa4ecb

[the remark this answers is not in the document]
erikw: Oh cool, I've never seen it before :)

README.md
```diff
@@ -25,13 +25,48 @@ Unfortunately restic does not come pre-configured with a way to run automated ba
...
 # Requirements
 * `restic >=v0.9.6`
 
-# Set up
+# TL;DR Setup
+1. Create B2 credentials as instructed [below](#1-create-backblaze-b2-account)
+1. Install config and scripts:
+   ```console
+   $ sudo make install
+   ```
```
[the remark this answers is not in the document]
gerardbosch: :+1: I added the original note for now, but if you feel a heading notice is more suitable no prob!

gerardbosch:
Update: I'm doing some final testing, I found something

gerardbosch:
Hey @erikw I've made some final testing on the whole thing together, and fixed a thing in the `$exclusion_args` as in my case I got sth like: `--exclude-file /home/user/.backup_exclude--exclude-file /home/user/.backup_exclude` (2 problems: missing space, and duplicated entry).

Now is ready, but I'd like to ask some observation I've just been thinking :)

1. The script traverses `/home/` looking for exclusion files for each user, BUT, will this only happen if at least `/home` is in `BACKUP_PATHS`? Will it be ignored if the path to backup is just `/` for example? And if so, should it ignore it? [that's not something I worry, as I only backup my homedir, but maybe nice to notice]

1. If all users in the system are checked for loading backup exclusions, it looks that one user could influence on the exclusions of another user, as for now, the ignore paths in `.backup_exclude` are absolute. So `user2` could define e.g.
   `/home/user2/.backup_exclude`:
   ```
   /home/user1/.config
   ```
   Would that be possible? If so, it would be maybe better opening an issue than trying to address it else here :smile: 

1. Finally, like in my case, if the only backup path is `/home/myuser`, the script generates the exclusion duplicate (I made a quick workaround for this): one per `.backup_exclude` in backup path, and another one for `.backup_exclude` in homedir.

   I'm not sure at all if is it worth it to improve on that or just the workaround using `sort unique` is enough on the long term :)

So that's it! Happy to contribute!! :smiley:

gerardbosch:
> Will restic complain on duplicates, or is it to trim down the command string? :)

Yep! I wrote it in the commit message body, but I quote it here, hehe.

> Having duplicate arguments seems to make no harm (other than probably
making restic parser do the job twice), but is weird and probably not
desirable.

I don't think it makes a major problem anyway, I've just removed that just in case using `sort -u`.

erikw:
> 2\. If all users in the system are checked for loading backup exclusions, it looks that one user could influence on the exclusions of another user, as for now, the ignore paths in `.backup_exclude` are absolute. So `user2` could define e.g.
>         `/home/user2/.backup_exclude`:

I think you hit a good point here. I realize now that this solution is aiming to solve a problem bigger than the scope of this project, multi-user backup.

It would make sense to slim this down  and make it simpler
* Always use `RESTIC_BACKUP_PATH`
*  Check for `.backup_EXCLUDE` in each path in `$BACKUP_PATHS` (for self-contained backup exclusions)

This project is anyways not aiming to be a out-of-the-box solution for everyone, but a base start to inspire hacking together a solution for everyone's unqiue needs :)

So while cool, the feature of automatic `/home/user*/.backup_exclude`inclusion, is bringing more trouble than use for this project scope. I think it can go :)

erikw:
> That will be no problem at all to me, as I'm doing single-user machine/homedir-only backup :D

✅ 2ba0864ae0b06cc44918b8cf95379968f7e3d3bb

erikw:
Alright, I just tested to use the new profiles and I successfully created a new incremental backup!

Are there any current open threads? I think we addressed everything in our discussion already @gerardbosch?

I would like to wait with merging this until after @mfeif has tested the setup from scratch as well. That will be the ultimate test :)

gerardbosch:
Hey @erikw, yes! I also think we addressed everything :D 

I've just pushed a function cleanup commit. I think that now is ready.

I confirm that I reinstalled yesterday everyghing `sudo make uninstall` and followed the TL;DR from README to check installation again. This morning when I unsuspended my laptop, the daily snapshot triggered and finished successfully :tada:  Let's see if @mfeif gives a double check :heavy_check_mark: so you can merge it :DD

This became a cool thread of collaboration :heart_decoration: , thanks for allowing this!

erikw:
Great, thanks for the clean-up. I missed that one.

I agree & enjoy it; thanks for the great collaboration all of you! I guess this is how GitHub was intended to be, social coding :)

mfeif:
Hey friends, I'm finally getting to testing this. Here are the questions/concerns I have in doing that:

1. The Makefile installs a `/usr/local/sbin/nm-unmetered-connection.sh` ... while I can see how that might be useful for folks, it seems out of scope for this utility. It doesn't seem like the restic scripts reference this script, so... ?
1. The _global.env file is gone? How can we cascade env variables without that? Therefore, the instructions in the Readme on the repo aren't correct; all of those files are not made by the Makefile.
1. The files have been named as .sh files, though they are not .sh scripts. This is nitpicking, but it might be confusing? (after running makefile, in /etc/restic I have: `b2_env.sh`, `b2_pw.txt`, `backup_exclude`)
1. Also, those files are installed via the Makefile, then the output from the Makefile says they're removed (why?) but they're still in my /etc/restic after finishing.
1. I like the starter kit for backup_exclude; those are sensible!
1. For systemd stuff, I end with `nm-unmetered-connection.service`, `restic-backup.service`, `restic-backup.timer`, `restic-check.service`, `restic-check.timer`, `status-email-user@.service` ... aren't those all supposed to have @ in them?

I think the lower-down non-tldr instructions might be more correct... I'll try them too, but I wanted to confine my comments to the "Makefile method" for this message.

mfeif:
I left a long comment with notes/comments, but it was on the wrong branch :-(

I'll start over on the right one!

gerardbosch:
> Hi @erikw, if we move ahead with this PR I would suggest to go with "squash-merge" as the log became a bit convoluted. Doing this the PR will be still co-authored I think :) Cheers!

Hi there! Just to remind this before merging, so the log in master doesn't get polluted :)

erikw:
> I left a long comment with notes/comments, but it was on the wrong branch :-(
> 
> I'll start over on the right one!

Thanks for trying out the PR!


You can also just link to the branch where you wrote your comment here? That's also fine :)

mfeif:
No, it was the master branch on this repo, which doesn't include all the relevant changes. My notes will come in my next post in a little bit...

mfeif:
Hey friends, I'm finally getting to testing this. Here are the questions/concerns I have found while doing that.

First, I tried following the makefile instructions:

1. The Makefile installs a `/usr/local/sbin/nm-unmetered-connection.sh` ... while I can see how that might be useful for folks, it seems out of scope for this utility. It doesn't seem like the restic scripts reference this script, but the .service files DO reference it, in fact, they depend on it. 
1. That script works with NetworkManager.service, which is not installed by default on a server, for example, Debian 11 doesn't have it.
1. Ditto for `/usr/local/sbin/systemd-email`; I actually am grateful for that script, but is it about backups? This is a nitpick.
1. The output of the makefile has some confusing stuff about installing things, then installing them again, and then rm'ing them:
    ```
    install -m 0600 etc/restic/_global.env.template etc/restic/_global.env
    install -m 0600 etc/restic/default.env.template etc/restic/default.env
    install -m 0600 etc/restic/pw.txt.template etc/restic/pw.txt
    install -d /etc/restic
    install -b -m 0600 etc/restic/_global.env etc/restic/backup_exclude etc/restic/default.env etc/restic/pw.txt /etc/restic
    rm -f etc/restic/_global.env etc/restic/default.env etc/restic/pw.txt
    ```
    It looks like there are two stanzas in the Makefile that do that; one is tagged the `install-conf` target, the other is just kinda hovering in the file.

5. I like the starter kit for backup_exclude; those are sensible!
6. In the Makefile, $(RM) is referenced, but not defined (maybe we get that for free from make?); maybe this falls back to just whatever rm is on the path; not sure. But also, I'm not sure why the script is rm'ing those (or attempting to)
7. In my previous hacky version, there was also a way to have exclude definitions in the `default.env`... that seems cleaner than having path/target specific excludes in the global.env. Can we get that back?
8. On step 4 (which links to the other step 3)... don't we have to source the _global.env first to be the same as the .service runner?
9. In step 5, perhaps explain here whether the user needs to make symlinks, or if they're automatically created by systemd (or makefile?)
10. In step 8, ditto about sourcing _global.env first?
11. Regarding the timers; perhaps some prose about how as-is, all the timers that are symlinks that reference the same source file will fire at the same time (I think?); that's probably not what you'd want if you had say 4-5 targets.

The script file itself seems to work great! My test backup is showing up in B2...

That's a lot there; I'll clean this stuff off the server and go through the manual config notes next, but in case we want to talk about any of the above first, I'll stop here for the night.

erikw:
Really great feedback @mfeif! I add my replies inline below

> * The Makefile installs a `/usr/local/sbin/nm-unmetered-connection.sh` ... while I can see how that might be useful for folks, it seems out of scope for this utility. It doesn't seem like the restic scripts reference this script, but the .service files DO reference it, in fact, they depend on it.
> * That script works with NetworkManager.service, which is not installed by default on a server, for example, Debian 11 doesn't have it.

The addition of the unmeetered check comes from #17. I would agree that this is an optional feature and should not be installed by default. There could possibly be a separate make target like `install-unmerted` or something, or just manual. To keep the scope of this already quite large PR down, I created a follow-up issue for this at #58 


 
> * Ditto for `/usr/local/sbin/systemd-email`; I actually am grateful for that script, but is it about backups? This is a nitpick.

Same here, #58. I would like to include the scripts in the repo as example or starts of custom solution, but it need not be installed by default. You are right.
 
> * The output of the makefile has some confusing stuff about installing things, then installing them again, and then rm'ing them:
>   ```
>   install -m 0600 etc/restic/_global.env.template etc/restic/_global.env
>   install -m 0600 etc/restic/default.env.template etc/restic/default.env
>   install -m 0600 etc/restic/pw.txt.template etc/restic/pw.txt
>   install -d /etc/restic
>   install -b -m 0600 etc/restic/_global.env etc/restic/backup_exclude etc/restic/default.env etc/restic/pw.txt /etc/restic
>   rm -f etc/restic/_global.env etc/restic/default.env etc/restic/pw.txt
>   ```

>   It looks like there are two stanzas in the Makefile that do that; one is tagged the `install-conf` target, the other is just kinda hovering in the file.

No worries about this. The Makefile does what is intended, but it does not matter because I will completely re-write the Makefile to solve #49 after this PR is merged. I have not decided exactly yet for now, but I think I will try to create a local `build/` directory, modify the PREFIX paths, and then install all files in the `build/` directory. Well have to see. But this will be addreseed after this PR is merged.

> * In the Makefile, $(RM) is referenced, but not defined (maybe we get that for free from make?); maybe this falls back to just whatever rm is on the path; not sure. But also, I'm not sure why the script is rm'ing those (or attempting to)

Yes you are correct. `RM` is set by make and defaults to `rm -f`. See https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

This whole thinkg about copy and rm during install comes from #15 (check for background). However this will also be made in a different way when I fix #49. So no need to worry about this for now :)

> * In my previous hacky version, there was also a way to have exclude definitions in the `default.env`... that seems cleaner than having path/target specific excludes in the global.env. Can we get that back?

Hmm could you maybe elaborate on what you mean by definitions? I'm not sure I follow :). 

Here's a comparision between `default.env`
* from the original PR (#14): https://github.com/erikw/restic-systemd-automatic-backup/blob/e186cc3b29a5c9435450d7f0f518df2f86f1731e/etc/restic/_global.env
* with PR: https://github.com/erikw/restic-systemd-automatic-backup/blob/ef41587f84ce65a68325afa70641ae6411b7707a/etc/restic/_global.env.template

To me, it looks like they both provide the same features and capabilities.

> * On step 4 (which links to the other step 3)... don't we have to source the _global.env first to be the same as the .service runner?

`default.env`  will source `_global.env` :). See
https://github.com/erikw/restic-systemd-automatic-backup/blob/ef41587f84ce65a68325afa70641ae6411b7707a/etc/restic/default.env.template#L12

> * In step 5, perhaps explain here whether the user needs to make symlinks, or if they're automatically created by systemd (or makefile?)

`$ systemctl enable` creates the symlinks for automatically starting the service. I added a short note in the README step :)

> * Regarding the timers; perhaps some prose about how as-is, all the timers that are symlinks that reference the same source file will fire at the same time (I think?); that's probably not what you'd want if you had say 4-5 targets.

Hmm that the targets use the same script (`restic_backup.sh`) should not be an issue. The system load could be a bit high if all timers are started at the same time though. That is a good point!

I added a note in README about editing this value.

To solve the problem with multiple backup profiles firing off at the same time, we can sleep a random delay on each execution. This is a common way to solve the problem in cronjobs.

We can update `restic-backup.service` to

```diff
- ExecStart=bash -c 'source /etc/restic/%I.env && /usr/local/sbin/restic_backup.sh | systemd-cat'
+ ExecStart=bash -c 'sleep ${RANDOM:0:3600} && source /etc/restic/%I.env && /usr/local/sbin/restic_backup.sh | systemd-cat'
```

This would make it sleep some random time between 0 to 60 minutes.

@mfeif @gerardbosch sounds good? :)

> The script file itself seems to work great! My test backup is showing up in B2...

This is the most important thing after all :check: :D

README.md
```diff
@@ -44,7 +44,9 @@ Note, you can use any of the supported [storage backends](https://restic.readthe
    * `/etc/restic/default.env` - Profile specific environment variables (multiple profiles can be defined by copying to `/etc/restic/something.env`).
    * `/etc/restic/backup_exclude` - List of file patterns to ignore. This will trim down your backup size and the speed of the backup a lot when done properly!
 1. Initialize remote repo as described [below](#3-initialize-remote-repo)
-1. Enable automated backup for starting with the system:
+1. Configure [how often]https://www.freedesktop.org/software/systemd/man/systemd.time.html#Calendar%20Events() back up should be made.
```
gerardbosch, line 47: Shouldn't `[how often]https://...()` become `[how often](https://...)`? ;)

README.md
```diff
@@ -158,6 +160,7 @@ Put these files in `/etc/systemd/system/`:
 * `restic-backup@.service`: A service that calls the backup script with the specified profile. The profile is specified
   by the value after `@` when running it (see below).
 * `restic-backup@.timer`: A timer that starts the former backup every day (same thing about profile here).
+   * If needed, edit this file to configure [how often]https://www.freedesktop.org/software/systemd/man/systemd.time.html#Calendar%20Events() back up should be made. See the `OnCalendar` key in the file.
```
gerardbosch, line 163: Idem: `[]URL()` vs. `[](URL)`

README.md
erikw: Yesyes! I've not learned to work properly with the new auto-close plugin in my editor that I use yet. thx!

erikw:
Sure! A user of multiple profiles can easily adjust the value anyways to optimize for their system.

The best solution would of course be something like a lock file (if load or bandwidth is a problem), but the complexity of this outweighs the benefits. For now, a 5 minute random sleep should suffice!

I'll push this in a moment.


Apart from that, I think the only remaining question was 

> > in my previous hacky version, there was also a way to have exclude definitions in the `default.env`... that seems cleaner than having path/target specific excludes in the global.env. Can we get that back?
> 
> Hmm could you maybe elaborate on what you mean by definitions? I'm not sure I follow :).
> 
> Here's a comparison between `default.env`
> 
>     * from the original PR (

Then we can soon finally merge this!

etc/systemd/system/restic-backup@.service
```diff
@@ -8,5 +8,6 @@ Type=simple
 Nice=10
 # $HOME or $XDG_CACHE_HOME must be set for restic to find /root/.cache/restic/
 Environment="HOME=/root"
+# The random sleep (in seconds) is in the case of multiple backup profiles. Many restic instances started at the same time could case high load or network bandwith usage.
 # `systemd-cat` allows showing the restic output to the systemd journal
-ExecStart=bash -c 'source /etc/restic/%I.env && /usr/local/sbin/restic_backup.sh | systemd-cat'
+ExecStart=bash -c 'sleep ${RANDOM:0:300} && source /etc/restic/%I.env && /usr/local/sbin/restic_backup.sh | systemd-cat'
```
gerardbosch, line 13: I didn't know the existence of the built-in `$RANDOM`. But the bounds: `:0:300` does not seem to work in this way.

Using `${VAR:x:y}` I think it is taken as a string, where `x` is the starting position, and optionally `y` the length. I run it and I got this:

```console
❯ echo ${RANDOM:0:300}
22519

❯ echo ${RANDOM:0:1}
7
```

gerardbosch: This seems to work:
`echo $((RANDOM % 300))` 
or
` echo $RANDOM % 10 + 1 | bc`

erikw: Thanks for checking!

mfeif:
Great stuff, guys. I will do the manual instructions following today.

Thanks for addressing all of my questions point by point! 

> Apart from that, I think the only remaining question was
> 
> > > in my previous hacky version, there was also a way to have exclude definitions in the `default.env`... that seems cleaner than having path/target specific excludes in the global.env. Can we get that back?
> > 
> > 
> > Here's a comparison between `default.env`
> > 
> > * from the original PR ([Multi repos, pull env variables out of scripts #14](https://github.com/erikw/restic-systemd-automatic-backup/pull/14)): https://github.com/erikw/restic-systemd-automatic-backup/blob/e186cc3b29a5c9435450d7f0f518df2f86f1731e/etc/restic/_global.env
> > * with PR: https://github.com/erikw/restic-systemd-automatic-backup/blob/ef41587f84ce65a68325afa70641ae6411b7707a/etc/restic/_global.env.template
> > 
> > To me, it looks like they both provide the same features and capabilities.

I agree that's the only thing; sorry I wasn't more clear. The comparision there is actually _global.env, not default.env, but really I just need to explain myself better: in my old version [linked here; inspect the example.env file](https://github.com/erikw/restic-systemd-automatic-backup/pull/14/files)... we had *local specific excludes*:

```
BACKUP_EXCLUDES="--exclude-file /.backup_exclude --exclude-file /mnt/media/.backup_exclude --exclude-file /home/erikw/.backup_exclude"
```

If I'm reading the current set of files, one can only put excludes into _global.env... it seems like it would be common that different backup plans/targets might have their own exclude setups, and putting them all in _global kinda removes the point of even having the cascading/inheriting behavior. I'd recommend we put those back, and make sure that the script uses BOTH excludes from _global and specific .env files

gerardbosch:
Hi @mfeif if I'm understanding you fine, that's something already possible. Check if this comment helps: https://github.com/erikw/restic-systemd-automatic-backup/pull/45#issuecomment-1015446697

The thing is that you can override any variable defined in `_global.env` in any specific profile -this includes the `$BACKUP_EXCLUDE`. Does it make sense?

EDIT: Oh! I think I get what you mean! You want to be able to define multiple exclusion files, is it? I've seen somewhere in the repo something like `$RESTIC_EXTRA_ARGS`, that I think it could help (maybe it is a new feature not merged, but I've seen it somewhere, hehe). For sure @erikw can enlighten :)

erikw:
@gerardbosch exactly, I will add extra args feature here #56

mfeif:
Yeah, that's what I mean.

Looking at [restic docs about excludes](https://restic.readthedocs.io/en/stable/040_backup.html#excluding-files) there are a number of ways to add excludes; if we want to keep that as files (rather than explicit mentions) we can pass more than one exclude file:

> --exclude-file Specified one or more times to exclude items listed in a given file

I might suggest that the script just look in both places (specific and global) for `RESTIC_BACKUP_EXCLUDE=` declarations and pass multiple --exclude-file params to the ultimate restic command.

I would also suggest that we change that variable to `RESTIC_BACKUP_EXCLUDES_FILE` or even `ADDITIONAL_EXCLUDES_FILE`

Thanks!

erikw:
> I would also suggest that we change that variable to `RESTIC_BACKUP_EXCLUDES_FILE` or even `ADDITIONAL_EXCLUDES_FILE`

Good idea. Done ✅ 

After #56, one can for example do in `some_profile.env`

```bash
# Complete override of global exclude file
RESTIC_BACKUP_EXCLUDES_FILE=/different/exclude/file

# Or add extra exclude files
export RESTIC_BACKUP_EXTRA_ARGS="--exclude-file  /exclude/file/a --exclude-file /exclude/file/b"
```

mfeif:
I'm not sure we need to use the --extra-args feature for my idea, since restic can take multiple `--exclude-file` arguments. Though I can imagine there will be other custom things that people want, so it's a good idea.

Would it be hard to modify the script to look for another env variable `ADDITIONAL_EXCLUDES_FILE` and add another --exclude-file parameter, just like _global? Splitting hairs here, I know, but I like the idea of it working the same way in both cases.

mfeif:
These comments are about the second set of instructions in the readme; the detailed ones:

* In step 9, what if we change:

   > There are some * check *-files in this git repo. Install these in the same way you installed the * -backup *-files and enable with sytemd

  to

  ```
  There are companion scripts to restic-backup.sh for checking; look in the repo in `etc/systemd/system` and `usr/local/sbin` and copy what you need over to their corresponding locations
  ```

erikw:
> I'm not sure we need to use the --extra-args feature for my idea, since restic can take multiple `--exclude-file` arguments. Though I can imagine there will be other custom things that people want, so it's a good idea.
> 
> Would it be hard to modify the script to look for another env variable `ADDITIONAL_EXCLUDES_FILE` and add another --exclude-file parameter, just like _global? Splitting hairs here, I know, but I like the idea of it working the same way in both cases.

I see what you mean. It could certainly be done :). The aim of the project is to keep it simple and extensible, so have and EXTRA_ARGS would be more general and cover more cases, so let's go with this for now.

I'll get started with #56 asap!

> * In step 2, we better mention `_global.env` and the starter `backup_exclude` file.

✅ good point, fixed!
 
> * In step 4, while it's true that a user can modify the script, I think our instructions on how to use the .env files would be better than the below:
>   > restic_backup.sh: A script that defines how to run the backup. Edit this file to respect your needs in terms of backup which paths to backup, retention (number of backups to save), etc.

✅ good catch again. Fixed!

> * In step 9, what if we change:
>   > There are some * check *-files in this git repo. Install these in the same way you installed the * -backup *-files and enable with sytemd
>   
>   
>   to
✅ sounds better, I inserted your suggestion

erikw:
aaaaand merged!

Thanks a lot @mfeif & @gerardbosch for your collaboration :)

I will update CHANGELOG, make a new release and also update the arch PKGBUILD in a bit

gerardbosch:
Nice :) You're welcome!
I think I could show a couple of ideas/PoC to expand on usability or features in the next days, let's see

PR #48, opened 2022-01-16
Fix make install failing when run multiple times

### LINKED ISSUE #46: make install twice ends in error
Hi there! :wave:

I think the following problem was introduced in https://github.com/erikw/restic-systemd-automatic-backup/pull/41/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R6

or maybe the root cause comes from here https://github.com/erikw/restic-systemd-automatic-backup/pull/15

Run `sudo make install` and again `sudo make install`. The second time you run it you get:

```console
install -d /tmp/a/b/usr/local/sbin
install -m 0744 usr/local/sbin/nm-unmetered-connection.sh usr/local/sbin/restic_backup.sh usr/local/sbin/restic_check.sh usr/local/sbin/systemd-email /tmp/a/b/usr/local/sbin
install -d /tmp/a/b/etc/restic
install -m 0600 etc/restic/b2_env.sh  etc/restic/b2_env.sh etc/restic/b2_pw.txt  etc/restic/b2_pw.txt etc/restic/backup_exclude /tmp/a/b/etc/restic
install: will not overwrite just-created '/tmp/a/b/etc/restic/b2_env.sh' with 'etc/restic/b2_env.sh'
install: will not overwrite just-created '/tmp/a/b/etc/restic/b2_pw.txt' with 'etc/restic/b2_pw.txt'
make: *** [Makefile:45: install-conf] Error 1
```

This is because once `make install` has run once, `SRCS_CONF = $(patsubst %.template, %, $(wildcard etc/restic/*))` expands to duplicated file names that fail on `cp/install` command, as 2 input files with the same name cannot be copied into the same directory –actually only a single file exists, but the macro seems to expand wrongly as consider both `%.template` and `*`.

---

After looking at this, I'm not very clear what's the purpose of copying (and then keeping) the `*.template` files into repository's directory: `etc/restic/b2_env.sh` and `etc/restic/b2_pwd.txt`, as these files are already installed into the system directories.

What do you think is the best way to fix this? I guess one option would be to get rid of the  non-template files in `etc/restic`, but I'm not sure if that was introduced on purpose or as a tactical thing to fix something [here](https://github.com/erikw/restic-systemd-automatic-backup/pull/41/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R38) :)

erikw:
Due to the security feature preventing checking in B2 credentials to
git, the install target would get the same file twice due to pattern
subsitution.

This issue is not visible with BSD install as it ignores the issue, but
GNU install wil fail.

Fixes #46

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,8 @@
 
 ### Macros ###
 SRCS_SCRIPTS	= $(filter-out %cron_mail, $(wildcard usr/local/sbin/*))
-SRCS_CONF	= $(patsubst %.template, %, $(wildcard etc/restic/*))
+# $(sort) remove duplicates that comes from running make install >1 times.
+SRCS_CONF	= $(sort $(patsubst %.template, %, $(wildcard etc/restic/*)))
 SRCS_SYSTEMD	= $(wildcard etc/systemd/system/*)
 
 # Just set PREFIX in environment, like


gerardbosch:
Nice! I'll get back to you tomorrow when I can test it :)

gerardbosch:
I tested the branch and the issue is fixed, thank you! I think you can merge it :)

erikw:
Thank you for testing :)

Issue #58, opened 2022-01-31
Make unmetered connection script optional


erikw:
From discussion in #45 

Either:
* separate make target `install-unmetered`,
* or just describe manually how to use it

Don't install this script by default.

gerardbosch:
Hi! For any laptop backup user, the unmetered check comes very handy. I'd say to make it very visible to the README it becomes optional.

Just an idea: Maybe another way is to only install it (and the service dependency) if NetworkManager is available on the target system.

Issue #49, opened 2022-01-18
PREFIX=/usr/local tries to install to /usr/local/usr/local..


erikw:
By default, it should install to `/sbin/, so that `$ PREFIX=/usr/local make install` installs as expected to `/usr/local/sbin`

I'm working on a fix already, but will not make PR or merge until we've merged #45 

If PREFIX is set, we need to `sed(1)`  update all references inside the files as well so that e.g. we look at `$PREFIX/etc/restic`.
 (or just sed-set one global `$PREFIX` variable which is used everywhere?)

PR #60, opened 2022-02-01
Add resticw (restic wrapper) utility

gerardbosch:
The script provides a convenient way to load environment config, deal
with profiles and act as a pass-through to restic. The overall thing is
to improve the UX when running restic, integrating the features this
project provides.

## Note

The script itself is a very simple thing.
The command line parser is auto-generated using [docopt.sh](https://github.com/andsens/docopt.sh) driven from
the script's DOC. It can be refreshed upon DOC changes with:
`docopt.sh path/to/resticw`.

### Examples

```console
resticw stats latest

resticw -p profileA snapshots
```

(No need of sourcing any file, works on any spawned terminal)

### Help

```console
❯ resticw --help
A little wrapper over restic just to handle profiles and environment loading.

  It loads the backup profile/environment in a subshell to avoid any credential leak (Note: Run it with sudo so it can load the environment).

  Usage:
    resticw [options] <restic_arguments>

    The restic_arguments is just the regular unwrapped restic arguments, e.g. stats latest

  Options:
    -p --profile=<name>        Specify the profile to load or use default [default: default].

  Examples:
    resticw --profile profileA snapshots
    resticw stats latest  # this will use the profile: default
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,8 @@ install: install-scripts install-conf install-systemd
 # target: install-scripts - Install executables.
 install-scripts:
 	install -d $(DEST_SCRIPTS)
-	install -m 0744 $(SRCS_SCRIPTS) $(DEST_SCRIPTS)
+	install -m 0744 $(filter-out %/resticw, $(SRCS_SCRIPTS)) $(DEST_SCRIPTS)
+	install -m 0755 usr/local/sbin/resticw $(DEST_SCRIPTS)
 
 # Copy templates to new files with restricted permissions.
 # Why? Because the non-template files are git-ignored to prevent that someone who clones or forks this repo checks in their sensitive data like the B2 password!
--- a/README.md
+++ b/README.md
@@ -80,23 +80,6 @@ Nevertheless the project should work out of the box, be minimal but still open t
    ````
 1. (optional) Setup email on failure as described [here](#8-email-notification-on-failure)
 
-# 🏃 Restic wrapper
-
-For convenience there's a `restic` wrapper script that makes loading profiles and **running restic**
-straightforward (it needs to run with sudo to read environment). Just run:
-
-- `sudo resticw WHATEVER` (e.g. `sudo resticw snapshots`) to use the default profile.
-- You can run the wrapper by passing a specific profile: `resticw -p anotherprofile snapshots`.
-
-### Useful commands
-
-| Command                                           | Description                                                       |
-|---------------------------------------------------|-------------------------------------------------------------------|
-| `resticw snapshots`                               | List backup snapshots                                             |
-| `resticw diff <snapshot-id> latest`               | Show the changes from the latest backup                           |
-| `resticw stats` / `resticw stats snapshot-id ...` | Show the statistics for the whole repo or the specified snapshots |
-| `resticw mount /mnt/restic`                       | Mount your remote repository                                      |
-
 
 # Step-by-step and manual setup
 This is a more detailed explanation than the TL;DR section that will give you more understanding in the setup, and maybe inspire you to develop your own setup based on this one even!
@@ -241,6 +224,21 @@ $ systemctl start restic-check@default.timer
 $ systemctl enable restic-check@default.timer
 ````
 
+## 10. Optional: 🏃 Restic wrapper
+For convenience there's a `restic` wrapper script that makes loading profiles and **running restic**
+straightforward (it needs to run with sudo to read environment). Just run:
+
+- `sudo resticw WHATEVER` (e.g. `sudo resticw snapshots`) to use the default profile.
+- You can run the wrapper by passing a specific profile: `resticw -p anotherprofile snapshots`.
+
+### Useful commands
+
+| Command                                           | Description                                                       |
+|---------------------------------------------------|-------------------------------------------------------------------|
+| `resticw snapshots`                               | List backup snapshots                                             |
+| `resticw diff <snapshot-id> latest`               | Show the changes from the latest backup                           |
+| `resticw stats` / `resticw stats snapshot-id ...` | Show the statistics for the whole repo or the specified snapshots |
+| `resticw mount /mnt/restic`                       | Mount your remote repository                                      |
 
 
 # Cron?
--- a/README.md
+++ b/README.md
@@ -262,10 +262,15 @@ A list of variations of this setup:
 * Using `--files-from` [#44](https://github.com/erikw/restic-systemd-automatic-backup/issues/44)
 
 # Development
-To not mess up your real installation when changing the `Makefile` simply install to a `$PREFIX` like
-```console
-$ PREFIX=/tmp/restic-test make install
-```
+* To not mess up your real installation when changing the `Makefile` simply install to a `$PREFIX` like
+   ```console
+   $ PREFIX=/tmp/restic-test make install
+   ```
+* **Updating the `resticw` parser:** If you ever update the usage `DOC`, you will need to refresh the auto-generated parser:
+  ```console
+  $ pip install doctopt.sh
+  $ doctopt.sh usr/local/sbin/resticw
+  ```
 
 # Releasing
 To make a new release:
--- a/README.md
+++ b/README.md
@@ -80,6 +80,24 @@ Nevertheless the project should work out of the box, be minimal but still open t
    ````
 1. (optional) Setup email on failure as described [here](#8-email-notification-on-failure)
 
+# 🏃 Restic wrapper
+
+For convenience there's a `restic` wrapper script that makes loading profiles and **running restic**
+straightforward (it needs to run with sudo to read environment). Just run:
+
+- `sudo resticw WHATEVER` (e.g. `sudo resticw snapshots`) to use the default profile.
+- You can run the wrapper by passing a specific profile: `resticw -p anotherprofile snapshots`.
+
+### Useful commands
+
+| Command                                           | Description                                                       |
+|---------------------------------------------------|-------------------------------------------------------------------|
+| `resticw snapshots`                               | List backup snapshots                                             |
+| `resticw diff <snapshot-id> latest`               | Show the changes from the latest backup                           |
+| `resticw stats` / `resticw stats snapshot-id ...` | Show the statistics for the whole repo or the specified snapshots |
+| `resticw mount /mnt/restic`                       | Mount your remote repository                                      |
+
+
 # Step-by-step and manual setup
 This is a more detailed explanation than the TL;DR section that will give you more understanding in the setup, and maybe inspire you to develop your own setup based on this one even!
 
--- a/usr/local/sbin/resticw
+++ b/usr/local/sbin/resticw
@@ -143,5 +143,5 @@ fi
 
 echo -e "‣ Using profile: ${__profile}  --  (${profile_file})\n"
 
-# shellcheck disable=SC2154
+# shellcheck disable=SC2154,SC1090
 source "$profile_file" && restic "$_restic_arguments_line_"
--- a/usr/local/sbin/resticw
+++ b/usr/local/sbin/resticw
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-# Exit on error, unbound variable, pipe error
-set -euo pipefail
 
 DOC="A little wrapper over restic just to handle profiles and environment loading.
 
@@ -120,12 +118,14 @@ declare -p "${prefix}__profile" "${prefix}_restic_arguments_line_"; done; }
 # docopt parser above, complete command for generating this parser is `docopt.sh resticw`
 
 # Parse arguments
-eval "$(set +u; docopt "$@")" # See https://github.com/andsens/docopt.sh for the magic :)
+eval "$(docopt "$@")" # See https://github.com/andsens/docopt.sh for the magic :)
 
 # --^^^-- END OF GENERATED COMMAND LINE PARSING STUFF --^^^--
 #
 # --vvv--             ACTUAL SCRIPT BELOW             --vvv--
 
+# Exit on error, unbound variable, pipe error
+set -euo pipefail
 ENV_DIR=/etc/restic
 
 ERR_NO_SUCH_PROFILE=2
@@ -137,7 +137,7 @@ profile_file="${ENV_DIR}/${__profile}.env"
 [[ ! -f "$profile_file" ]] && echo "Invalid profile: No such environment file ${profile_file}" && exit "$ERR_NO_SUCH_PROFILE"
 
 if [[ ! -r "$profile_file" ]]; then
-  echo "Error: Run it as root or use sudo, as the script needs to load the environment from protected file ${profile_file}"
+  echo "Error: could not read the environment file ${profile_file}. Are you running this script as the correct user? Maybe try sudo with the right user.
   exit "$ERR_PROFILE_NO_READ_PERM"
 fi
 
--- a/usr/local/sbin/resticw
+++ b/usr/local/sbin/resticw
@@ -130,13 +130,16 @@ eval "$(set +u; docopt "$@")" # See https://github.com/andsens/docopt.sh for the
 
 ENV_DIR=/etc/restic
 
+ERR_NO_SUCH_PROFILE=2
+ERR_PROFILE_NO_READ_PERM=3
+
 profile_file="${ENV_DIR}/${__profile}.env"
 
-[[ ! -f "$profile_file" ]] && echo "Invalid profile: No such environment file ${profile_file}" && exit 2
+[[ ! -f "$profile_file" ]] && echo "Invalid profile: No such environment file ${profile_file}" && exit "$ERR_NO_SUCH_PROFILE"
 
-if [ $(id -u) -ne 0 ]; then
+if [[ ! -r "$profile_file" ]]; then
   echo "Error: Run it as root or use sudo, as the script needs to load the environment from protected file ${profile_file}"
-  exit 3
+  exit "$ERR_PROFILE_NO_READ_PERM"
 fi
 
 echo -e "‣ Using profile: ${__profile}  --  (${profile_file})\n"
--- a/usr/local/sbin/resticw
+++ b/usr/local/sbin/resticw
@@ -5,8 +5,6 @@ set -euo pipefail
 
 DOC="A little wrapper over restic just to handle profiles and environment loading.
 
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -31,7 +31,8 @@ install: install-scripts install-conf install-systemd
 # target: install-scripts - Install executables.
 install-scripts:
 	install -d $(DEST_SCRIPTS)
-	install -m 0744 $(SRCS_SCRIPTS) $(DEST_SCRIPTS)
+	install -m 0744 $(filter-out %/resticw, $(SRCS_SCRIPTS)) $(DEST_SCRIPTS)
+	install -m 0755 usr/local/sbin/resticw $(DEST_SCRIPTS)
```
gerardbosch, line 35: Using 755 here as, the user should be able to find `resticw` discoverable, but by using 744, `which resticw` shows nothing, as well as tab  auto-completion does not help when typing the command.

gerardbosch:
The parser could scare, but as it is auto-generated it does not require the burden of manual manteinance, just issue a command. I always found CLI parsing tough, hehe, and found docopt.sh very nice.

erikw:
I just sneak-peaked a bit and it looks interesting!

I make a proper review and try-out later in the week :)

README.md
```diff
@@ -75,6 +75,24 @@ Note, you can use any of the supported [storage backends](https://restic.readthe
    ````
 1. (optional) Setup email on failure as described [here](#8-email-notification-on-failure)
 
+# 🏃 Restic wrapper
```
erikw, line 78: Thanks for writing REAME docs ⭐ 

To keep the base install minimal,  simple, easy to follow, let's move this to after "9. Optional: automated backup checks" as "10. Optional: restic profile wrapper" or something like this.

Later in  #58 I will restructure the optionals better in the README

gerardbosch:
Sure! There is one part of the update mechanism implemented as a separate bash script, as Makefile started to get convoluted and limitating.

These changes will require adjustments as per the recently added linter. But I keep it on hold for a while :)

README.md
```diff
@@ -75,6 +75,24 @@ Note, you can use any of the supported [storage backends](https://restic.readthe
...
+### Useful commands
+
+| Command                                           | Description                                                       |
+|---------------------------------------------------|-------------------------------------------------------------------|
+| `resticw snapshots`                               | List backup snapshots                                             |
+| `resticw diff <snapshot-id> latest`               | Show the changes from the latest backup                           |
+| `resticw stats` / `resticw stats snapshot-id ...` | Show the statistics for the whole repo or the specified snapshots |
+| `resticw mount /mnt/restic`                       | Mount your remote repository                                      |
+
+
```
erikw, line 95: To make this easy to maintain, let's add a bullet point in the section `# Development`, that lists the minimal commands to update the parser

```console
$ pip install doctopt.sh
$ doctopt.sh usr/local/sbin/resticw
```

or something. It's very useful to have these short command notes down the line when adding a new paramter in a few years.. :)

usr/local/sbin/resticw
```diff
@@ -0,0 +1,141 @@
...
+    sudo resticw --profile profileA snapshots
+    sudo resticw stats latest  # this will use the profile: default
+"
+
+# The following argument parser is generated with docopt.sh from the above docstring.
+# See https://github.com/andsens/docopt.sh. If the DOC is updated or new options are added, refresh the parser!
+
+# docopt parser below, refresh this parser with `docopt.sh resticw`
+# shellcheck disable=2016,1075
+docopt() { parse() { if ${DOCOPT_DOC_CHECK:-true}; then local doc_hash
```
erikw, line 28: I think docopt looks cook, but it's sure overkill for this use case of having just one option. Just checking manually `$1`, or using the standard built-in `getopts` (see my template here https://github.com/erikw/templates/blob/master/bash/getopts.sh#L30-L38) would suffice.

But programming should also be fun, and trying out new techs is fun,... so let's give it a try :)

usr/local/sbin/resticw
```diff
@@ -0,0 +1,141 @@
...
+
+# --^^^-- END OF GENERATED COMMAND LINE PARSING STUFF --^^^--
+#
+# --vvv--             ACTUAL SCRIPT BELOW             --vvv--
+
+ENV_DIR=/etc/restic
+
+profile_file="${ENV_DIR}/${__profile}.env"
+
+[[ ! -f "$profile_file" ]] && echo "Invalid profile: No such environment file ${profile_file}" && exit 2
```
erikw, line 131: 👍 for unique exit codes

usr/local/sbin/resticw
```diff
@@ -0,0 +1,141 @@
...
+
+if [ $(id -u) -ne 0 ]; then
```
erikw, line 133: Hmm I would disagree actually, this project need actually not be run by root. If one installs it in a directory hierarchy owned by root, then yes.

For the macOS setup that will be added, the project will typically be installed in `/usr/local/` where normal users have write access (this is where the Homebrew package manager installed applications by default). This is why #49 is important to fix.

So I'd say let's remove this actually. It should be obvious if one is not running as the right user

usr/local/sbin/resticw
```diff
@@ -0,0 +1,141 @@
...
+  echo "Error: Run it as root or use sudo, as the script needs to load the environment from protected file ${profile_file}"
+  exit 3
+fi
+
+echo -e "‣ Using profile: ${__profile}  --  (${profile_file})\n"
```
erikw, line 138: Tip: use `printf(1)` for more advanced formatting.

```bash
printf "‣ Using profile: %s -- (%s)\n" "${__profile}" "${profile_file}"
```

usr/local/sbin/resticw
```diff
@@ -0,0 +1,141 @@
...
+
+# Source it in a subshell (avoid any credential leak), then run restic
```
erikw, line 140: How would credentials be leaked if we don't spawn a subshell? The script itself already runs in a sub-shell from where it was started.

erikw (changes requested):
Cool, I think this looks useful!

I've not actually tried it yet, to be done...

usr/local/sbin/resticw
```diff
@@ -0,0 +1,141 @@
...
+"${prefix}_restic_arguments_"
+eval "${prefix}"'__profile=${var___profile:-default}'
+eval "${prefix}"'_restic_arguments_=${var__restic_arguments_:-}'
+local docopt_i=1; [[ $BASH_VERSION =~ ^4.3 ]] && docopt_i=2
+for ((;docopt_i>0;docopt_i--)); do declare -p "${prefix}__profile" \
+"${prefix}_restic_arguments_"; done; }
+# docopt parser above, complete command for generating this parser is `docopt.sh resticw`
+
+# Parse arguments
+eval "$(set +u; docopt "$@")" # See https://github.com/andsens/docopt.sh for the magic :)
```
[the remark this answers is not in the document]
erikw: When I run `$ docopt.sh resticw` I get a warning:

```
resticw No invocations of docopt found, check your script to make sure this is correct.
docopt.sh is invoked with `eval "$(docopt "$@")"`.

```

README.md
```diff
@@ -75,6 +75,24 @@ Note, you can use any of the supported [storage backends](https://restic.readthe
    ````
 1. (optional) Setup email on failure as described [here](#8-email-notification-on-failure)
 
+# 🏃 Restic wrapper
+
+For convenience there's a `restic` wrapper script that makes loading profiles and **running restic**
+straightforward (it needs to run with sudo to read environment). Just run:
+
+- `sudo resticw WHATEVER` (e.g. `sudo resticw snapshots`) to use the default profile.
```
erikw, line 83: sudo not required as per comment below

usr/local/sbin/resticw
```diff
@@ -0,0 +1,141 @@
...
+
+# Exit on error, unbound variable, pipe error
+set -euo pipefail
+
+DOC="A little wrapper over restic just to handle profiles and environment loading.
+
+  It loads the backup profile/environment in a subshell to avoid any credential leak (Note: Run it with sudo so it can load the environment).
+
+  Usage:
+    resticw [options] <restic_arguments>
```
erikw, line 11: When just running like `$ resticw` without arguments, the helptext is missing the 2 last characters from this line:

```
   Usage:
      resticw [options] <restic_arguemnt
                                        ^
```

erikw: I think restic calls the first level of argument commands, which can have arguments. In `$restic backup /` `backup` would be the command and `/` the argument to that command =>

```
 Usage:
    resticw [restic_options] <restic_command> [restic_command_options]
```

?

erikw:
I've tested the script now. All in all it's great - I will use it myself on my systems :)

Just a few comments above to look at

gerardbosch:
Cool! I try to check the comments, answer the ones I can or address it in a while :D

Makefile
gerardbosch: :+1:

usr/local/sbin/resticw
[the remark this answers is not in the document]
gerardbosch: I opened few days ago this https://github.com/andsens/docopt.sh/issues/19 regarding a problem with `set -u`. The author answered on that with the `set +u` trick and it's view on that :)

But as you've just faced, there is a weird warning now. I've just written again to the issue to notice that false warning/error.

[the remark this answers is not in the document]
gerardbosch: It looks like a false alarm, as it works fine. I see 3 options:
- Remove the `set -u` from the top of the script and get rid of the `set +u` workaround then.
- Accept and live with that false error when refreshing the parser 😆
- Wait to see if the message gets fixed in origin, maybe less probable

README.md
gerardbosch: Perfect! No worries :)

README.md
gerardbosch: Sure, done!

usr/local/sbin/resticw
gerardbosch: Yeah, hehe. I liked to save myself the burden of writing that tough stuff. And it comes with nice benefits, like short and long options, `--help` and usage out-of-the-box :D  Also good way to try stuff.

gerardbosch: Also forces you to write good usage DOC messages, hehe

usr/local/sbin/resticw
gerardbosch: Great! Thanks to point that.
I've change it to only issue that message (and error) in case the profile is not readable by the user. Check last version and let me know :)

In any case, of failure, maybe is better to echo a warning and exit, than assume the user remembers about internals and file permissions. Not showing a message would make it fail anyway but maybe less obvious what's the problem.

usr/local/sbin/resticw
gerardbosch: Hey, thanks! I can't spot much of a difference, the `printf` version being even longer. Do you want to change this?

I find the string interpolation in echo much easier to read than using `%format` and passing actual arguments at the end. It is not using any advanced formating actually, hehe.

usr/local/sbin/resticw
gerardbosch: Oh! totally true! I removed the note, and the sub-shell. I think I got confused the first time I wrote this script time ago :) Credentials are confined to the shell spawned by running `resticw` and the environment dies after command finishes.

README.md
gerardbosch: Updated

usr/local/sbin/resticw
gerardbosch: Yep! I realised the missing text on `$ resticw` before, and after comparing with the docopt.sh example, now, I managed to fix it. It should work if you update the branch :)

---
For the other thing of `<restic_command> [restic_command_options]` I tried to change it and just finally reworded as `<restic_arguments_line>...`. I think splitting in 2 (command, options) does not work well with the parser. But it help me to fix something, now there is `...` that was missing to allow multiple tokens. Also, from the shell standpoint, they are the arguments provided to the restic command (it depends on which perspective you observe, hehe).

But the ultimate reason is that it didn't work well.

usr/local/sbin/resticw
erikw: I don't say no to good documentation!

usr/local/sbin/resticw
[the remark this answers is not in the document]
erikw: With the motivation that the acutal scrip starts at line 128, I think it makes sense to move `set -euo pipefail` there actually, and let docopt do its thing before that

```
# --vvv--             ACTUAL SCRIPT BELOW             --vvv--

set -euo pipefail
ENV_DIR=/etc/restic
[...]
```

usr/local/sbin/resticw
erikw: Preferences of style :). I see benefits in both. No need to fix, just a tip

erikw:
> One thing to ask, I've seen most of the stuff in the project is using tabs instead of spaces. I wrote that one with space indentation, but I can commit to replace it by tabs if you prefer, to be consistent with the rest of the project. I'm not tab fan, but no problem to change it, hehe.

Good point. I was raised with hard tabs, coming from a background C with Linux kernel coding style convention... and I still prefer hard tabs to soft everyday ;P

In the end it really does not matter, as long as the code base is consistent. We should add the `.editorconfig` I saw in your other PR.

For consistency, let's use hard tabs. With `.editorconfig` in place this will be transparent anyways :)

usr/local/sbin/resticw
erikw: Now it looks better :)

usr/local/sbin/resticw
erikw: That's a great approach, I like it!

Nitpick; technically the error message shoud lnot say  `Error: Run it as root or use sudo` because maybe someone has set up a different use account `restic-account` that is used for backup. The message should be more generic like `Error: can't ready the profile. Are you running this script as the right user? Try su or sudo`

or something like this :)

usr/local/sbin/resticw
```diff
@@ -0,0 +1,145 @@
...
+# --^^^-- END OF GENERATED COMMAND LINE PARSING STUFF --^^^--
+#
+# --vvv--             ACTUAL SCRIPT BELOW             --vvv--
+
+ENV_DIR=/etc/restic
+
+ERR_NO_SUCH_PROFILE=2
+ERR_PROFILE_NO_READ_PERM=3
+
+profile_file="${ENV_DIR}/${__profile}.env"
```
erikw, line 134: Linter says

```
In /github/workspace/usr/local/sbin/resticw line 134:
profile_file="${ENV_DIR}/${__profile}.env"
                         ^----------^ SC2154 (warning): __profile is referenced but not assigned.
```

I think it's not able to parse the docopt function above. We might need to [ignore](https://github.com/koalaman/shellcheck/wiki/Ignore) this check here with

```
# shellcheck disable=SC2154
profile_file="${ENV_DIR}/${__profile}.env"
```

erikw:
Thanks for the fixes! I "fixed" the linter warnings and the only two small things would be

* `set -euo pipefail`
* possibly error message when can't reading profile

usr/local/sbin/resticw
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env bash
+
+# Exit on error, unbound variable, pipe error
+set -euo pipefail
```
erikw, line 4:
```suggestion
```

usr/local/sbin/resticw
```diff
@@ -0,0 +1,147 @@
...
+declare -p "${prefix}__profile" "${prefix}_restic_arguments_line_"; done; }
+# docopt parser above, complete command for generating this parser is `docopt.sh resticw`
+
+# Parse arguments
+eval "$(set +u; docopt "$@")" # See https://github.com/andsens/docopt.sh for the magic :)
+
+# --^^^-- END OF GENERATED COMMAND LINE PARSING STUFF --^^^--
+#
+# --vvv--             ACTUAL SCRIPT BELOW             --vvv--
+
```
erikw, line 128:
```suggestion

# Exit on error, unbound variable, pipe error
set -euo pipefail
```

usr/local/sbin/resticw
```diff
@@ -0,0 +1,147 @@
...
+eval "${prefix}"'__profile=${var___profile:-default}'
+if declare -p var__restic_arguments_line_ >/dev/null 2>&1; then
+eval "${prefix}"'_restic_arguments_line_=("${var__restic_arguments_line_[@]}")'
+else eval "${prefix}"'_restic_arguments_line_=()'; fi; local docopt_i=1
+[[ $BASH_VERSION =~ ^4.3 ]] && docopt_i=2; for ((;docopt_i>0;docopt_i--)); do
+declare -p "${prefix}__profile" "${prefix}_restic_arguments_line_"; done; }
+# docopt parser above, complete command for generating this parser is `docopt.sh resticw`
+
+# Parse arguments
+eval "$(set +u; docopt "$@")" # See https://github.com/andsens/docopt.sh for the magic :)
```
erikw, line 123:
```suggestion
eval "$(docopt "$@")" # See https://github.com/andsens/docopt.sh for the magic :)
```

usr/local/sbin/resticw
```diff
@@ -0,0 +1,147 @@
...
+ERR_NO_SUCH_PROFILE=2
+ERR_PROFILE_NO_READ_PERM=3
+
+# shellcheck disable=SC2154
+profile_file="${ENV_DIR}/${__profile}.env"
+
+[[ ! -f "$profile_file" ]] && echo "Invalid profile: No such environment file ${profile_file}" && exit "$ERR_NO_SUCH_PROFILE"
+
+if [[ ! -r "$profile_file" ]]; then
+  echo "Error: Run it as root or use sudo, as the script needs to load the environment from protected file ${profile_file}"
```
erikw, line 140:
```suggestion
  echo "Error: could not read the environment file ${profile_file}. Are you running as the correct use? Maybe try sudo with the right user.
```

usr/local/sbin/resticw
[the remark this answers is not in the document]
gerardbosch: :facepalm:  oh yes! I didn't think on that solution!! I updated it with the other suggestions

usr/local/sbin/resticw
gerardbosch: Sure! I commited this along with other suggestions ;)

erikw:
Alrigh, it's in 🚀 🚀 !

gerardbosch:
Nice!

PR #61, opened 2022-02-01
Feature/make upgrade

gerardbosch:
Add upgrade target to the Makefile

`make upgrade` will help to upgrade the installation with new features
and/or breaking configuration changes. The upgrade process will keep a
copy (backup) of config files and make a best effort to restore previous
values.

User manual action should be not required for default settings (as long as
existing env vars keep same name), and be minimal in the case of
customizations.

The upgrade mechanism could be tested to install #60.


etc/restic/_global.env.template
```diff
@@ -17,4 +17,4 @@ export B2_ACCOUNT_ID="<b2-account-id>"   # TODO fill with your account info
 export B2_ACCOUNT_KEY="<b2-account-key>" # TODO fill with your account info
 
 # How many network connections to set up to B2. Default is 5.
-export B2_CONNECTIONS=10
+export B2_CONNECTIONS="10"
```
gerardbosch, line 20: Quoting variables makes the parsing part easier.

erikw:
I've not checked the actual code changes, but the idea sounds good. But hold on just a minute before investing more in to the PR because ...

as part of #49, I will re-evaluate if a Makefile is still the best tool for the job here. It's likely that the project will continue with make, but just as I can't say for sure yet, let's wait with this until I've figured out #49 :)

gerardbosch:
[I previously placed the following in the wrong thread, I move it here :) ]

Sure! There is one part of the update mechanism implemented as a separate bash script, as Makefile started to get convoluted and limitating.

These changes will require adjustments as per the recently added linter. But I keep it on hold for a while :)

.editorconfig
```diff
@@ -0,0 +1,20 @@
+# EditorConfig helps developers define and maintain consistent coding styles
```
erikw, line 1: This file exist in the `master` branch already now, thus not needed for this PR

Makefile
```diff
@@ -2,6 +2,7 @@
 .PHONY: help install install-scripts install-conf install-systemd uninstall
 
 ### Macros ###
+DEFAULT_BRANCH = master
```
erikw, line 5: I think we can get this on-the-fly with

```bash
basename $(git symbolic-ref refs/remotes/origin/HEAD)
```

as this is subject to change. I've been thinking about renaming `master` to `main` on old repos as well but not yet made it.

Makefile
```diff
@@ -25,7 +27,7 @@ help:
 	@egrep "#\starget:" [Mm]akefile  | sed 's/\s-\s/\t\t\t/' | cut -d " " -f3- | sort -d
 
 # target: install - Install all files
-install: install-scripts install-conf install-systemd
+install: install-scripts install-conf install-systemd install-current-version-file
```
erikw, line 30: This file is not relevant when installed as an AUR package. But the PKG can do an `rm` on it before building the package maybe

Makefile
```diff
@@ -51,9 +53,70 @@ install-systemd:
...
 uninstall:
 	@for file in $(INSTALLED_FILES); do \
 			echo $(RM) $$file; \
 			$(RM) $$file; \
 	done
+
+# target: upgrade - Upgrade installation.
+# .ONESHELL runs the script in a single shell instead of a shell per line; so no semicolons, nor backslashes required :)
+.ONESHELL:
+upgrade:
```
erikw, line 70: As this is quite a lot of logic, I think it would do better in a script of its own, with the needed inputs as arguments to it.

_helper/post-upgrade-config.sh
```diff
@@ -0,0 +1,91 @@
...
+trim_kv_pair() {
+	line="$1"
+	# considers patterns: key="VALUE" and key+="VALUE"
+	echo "$(sed -E 's|.*(\b\w+\b\+?=".*").*|\1|' <<< "$line")"
+}
+
+store_env_vars() {
+	local config_file_backup="$1"
+
+	local kv_lines="$(filter_var_assignments "$config_file_backup")"
```
erikw, line 32: This could work for simple cases. We have to keep in mind that the env files are shell scripts and someone could do something like building up a value over a few statements and with different syntax like

```bash
RESTIC_BACKUP_PATHS="$HOME/dir"
if [...]; then
   RESTIC_BACKUP_PATHS+=" /something"
else
   RESTIC_BACKUP_PATHS="${RESTIC_BACKUP_PATHS} /other"
end
```

Correctly finding the values for all cases I think is quite hard

Maybe a safer alternative would be to 
1. source the `.env.sh` file
2. store the values now available in envvars like `$RESTIC_BACKUP_PATHS`
3. Start from a new clean config and insert those values

Hmm actually this is also not perfect, as a typical thing could be

```bash
if [ -d /mnt/some_disk ] ; then
   RESTIC_BACKUP_PATHS+=" /mnt/some_disk"
end
```
and the value depends on how the system looked at this time.

etc/restic/_global.env.template
erikw: Yes, but we can not rely on that users will adhere to a specific style, anything that is bash is allowed

erikw (commented):
It's a cool idea to save the config values. I need to point out that there are already two ways that config values are being preserved

* as the introduction of the `-d` flag to `install(1), all config files are backed up with a datetime suffix now: https://github.com/erikw/restic-systemd-automatic-backup/blob/7a1409ca03ded08c74d23e5e07a07c68fd9b6f41/Makefile#L128-L130
* When installed in Arch Linux from the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=restic-systemd-automatic-backup), config files are backed up automatically with the `.pacsave` extension

With the addition of `install -d`, the urge for an upgrade command is less I would say, than before.

My two main worries are that
* I think this is very hard to get right and work properly for all different possible env files, which will lead to a lot of
* maintenance / bug fixes that could be hard for me to manage. I don't want this project to become too complex, the [project scope](https://github.com/erikw/restic-systemd-automatic-backup/blob/master/README.md#project-scope) is to be a hackable start after all

I suspect that most users, will not do upgrade very often, as they create their custom solution based on this one

gerardbosch:
Hey @erikw, yes! I agree with the overall thing you mention, and also think that all of these would be very fragile on the long run and hard to maintain. It was mostly a PoC to experiment and see if a transparent-automated upgrade would be feasible :)

The idea was to just make a "best effort" to migrate configuration values for a default installation –no user customization on environment files, etc.

I think that we may close this. I'll just keep the branch as a PoC, as I don't think it wouldn't be maintainable for myself either :laughing: ; or at most I will use it as-is for my own experimentation, hehe.

But if you have a different ideas, or find a nice way to do the upgrade and config migration, we can re-open it, of course!

gerardbosch:
Hey @erikw, yes! I agree with the overall thing you mention, and also think that all of these would be very fragile on the long run and hard to maintain. It was mostly a PoC to experiment and see if a transparent-automated upgrade would be feasible :)

The idea was to just make a "best effort" to migrate configuration values for a default installation –no user customization on environment files, etc.

I think that we may close this. I'll just keep the branch as a PoC, as I don't think it wouldn't be maintainable for myself either :laughing: ; or at most I will use it as-is for my own experimentation, hehe.

But if you have a different ideas, or find a nice way to do the upgrade and config migration, we can re-open it, of course!

PR #70, opened 2022-02-08
Update README: Add notice on immediate backup

gerardbosch:


gerardbosch:
I add a note on the README, but I'm not very convinced with the random delay. Maybe we could use `ps` to see if the `restic_backup.sh` is running instead of delaying, because when you want to trigger an immediate backup like

```console
sudo systemctl start restic-backup@default
journalctl -f -u restic-check@default
```

you would expect that the backup starts moving on, but you may need to sit and wait up to 5 minutes to see something happening.

erikw:
You are right. It's unnecessary to sleep if there is no other restic running. I pushed 6c1d182e21643c0c3ceb48c2352ceca010775401 and will close this PR instead :)

Thanks!

gerardbosch:
Super!! :rocket:

PR #71, opened 2022-02-08
Fix resticw arguments line

gerardbosch:
It needs to be treated as an array so that complex commands like
`stats latest` work.

### DIFF
--- a/bin/resticw
+++ b/bin/resticw
@@ -136,11 +136,12 @@ profile_file="${ENV_DIR}/${__profile}.env.sh"
 [[ ! -f "$profile_file" ]] && echo "Invalid profile: No such environment file ${profile_file}" && exit "$ERR_NO_SUCH_PROFILE"
 
 if [[ ! -r "$profile_file" ]]; then
-	echo "Error: could not read the environment file ${profile_file}. Are you running this script as the correct user? Maybe try sudo with the right user."
+	echo "Error: Could not read the environment file ${profile_file}. Are you running this script as the correct user? Maybe try sudo with the right user."
 	exit "$ERR_PROFILE_NO_READ_PERM"
 fi
 
 echo -e "‣ Using profile: ${__profile}  --  (${profile_file})\n"
 
 # shellcheck disable=SC2154,SC1090
-source "$profile_file" && restic "$_restic_arguments_line_"
+source "$profile_file" && restic "${_restic_arguments_line_[@]}"
+


gerardbosch:
@erikw I also found a limitation, this is that restic commands that take options will not work, for example

```
❯ sudo resticw stats latest --cache-dir=/tmp/cache
Usage:
  resticw [options] <restic_arguments_line>...
```

It seems to be because of the generated parser tries to parse an option that is not for the `resticw` but for the underlying command. I'm not very sure if there's any way to fix it or maybe a tradeoff of using the wraper.

I mostly use the wrapper for simple things like `snapshots`, `mount`, `stats` or `diff snapshotA snapshotB`. For this cases it works fine.

erikw:
I see. Hm maybe an issue to report at  https://github.com/andsens/docopt.sh

gerardbosch:
Hi there! I followed the advice and asked here: https://github.com/andsens/docopt.sh/issues/20

It seems that there's the way to make it work with mixed options (resticw/restic), it always was there.

I see 2 options:
- (a) Leave it as is now, and sell the `resticw` with that little limitation. For the usually everyday commands like `snapshots`, `stats`, `diff`,... _maybe_ is not a very necessary to pass additional options. The user can always source the profile and rely on plain restic.

- (b) Change the usage to always require the `--`, and offer a full restic support. This way the commands would be  like:
  ````console
  sudo resticw -p profileX -- snapshots
  sudo restic -- stats latest
  ...
  sudo restic [options] -- <restic line>...
  
What do you think? I've got not strong opinion actually, maybe AS-IS is fine enough, but not sure.

erikw:
Hmm that's really unfortunate. But I understand why.

`resticw` is not really a wrapper if it does not support options of `restic`.

I guess we could look for "-p profile" in "$@" and remove those before passing to docopt.sh. However the purpose of benefit of doctopt is not really there anymore. 

I'd say the following alternatives are all equally good for me:
* Handle args ourselves like https://github.com/andsens/docopt.sh/issues/20#issuecomment-1033824160
* Maybe `getopts` is an option, but maybe the problem is the same as with doctopt, that how do we distingush between args for our wrapper and for the wrapped command? (Not tried it, maybe it works dunno)
* force `--` usage in `resticw`

gerardbosch:
Hey! I liked that idea that docopt.sh could offer the possibility do do `strict` order parsing. As an optional feature for docopt sounds it could be useful. I'll link your comment to the issue to see what @andsens thinks about it.

I like the GNU style, but for this particular case the BSD one seems helpful :)

gerardbosch:
The side problem of properly wrapping restic options should be already FIXED, docopt.sh already provided that support :sweat_smile: :tada: : https://github.com/erikw/restic-automatic-backup-scheduler/pull/85

Issue #73, opened 2022-02-09
Installation via Homebrew


erikw:
Following up from #51, create a Homebrew Formula in my tap https://github.com/erikw/homebrew-tap

gerardbosch:
Hi again :) More questions, hehe.
I've been playing with Nix package manager recently, which also works on both Linux and MacOS. I'm still evaluating it, but there are certain things that I find great, mainly the hash-based immutable dependencies.

Do you think it would make sense to try publishing into Nix? I have to say that I've got zero idea how to do that, but currently I'm installing all the system packages I can from there (it has more than 70,000 packages, although I think this includes the Perl CPAN ones, and other language libraries).

erikw:
Maybe that. I'm using Arch's AUR and Homebrew, that's why I've focused on them here

I've only read about nix but not try used it myself.

If it's easy to maintain, why not :)

erikw:
Work in progress at https://github.com/erikw/homebrew-tap/tree/formula_restic

Issue #51, opened 2022-01-18
Add support for macOS


erikw:
As discussed in #47, I already have it running on a mac system for several years. However as a cron job. 

* Add instructions on how to set up on a mac in `/usr/local/`. #49 is a blocker for this
* Create a proper solution using [LaunchAgents](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/ScheduledJobs.html#//apple_ref/doc/uid/10000172i-CH1-SW2)
* Integrate with the Makefile, maybe `$ make install-launchagent` ?
* Create a Homebrew Formula in my tap https://github.com/erikw/homebrew-tap

gerardbosch:
Just for curiosity :) Is there much benefit by using LaunchAgents over cron?

erikw:
I've not used LaunchAgent for timers yet (but I do for services), so I don't know yet. But I suspect that it will give nicer features just like systemd timers gives over cron.

https://apple.stackexchange.com/a/409334/197493

To be investigated :)

PR #74, opened 2022-02-09
Add macOS launchagent support

erikw:
Scoped down #51 and created #73

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
-- macos LaunchAgent support. Install with `make install-launchagent`.
+- macos LaunchAgent support. Install with `make install-launchagent` and activate with `make activate-launchagent`. See [README.md](README.md) for details.
 
 ## [5.0.0] - 2022-02-08
 ### Added


README.md
```diff
@@ -260,6 +260,36 @@ straightforward (it needs to run with sudo to read environment). Just run:
 | `resticw stats` / `resticw stats snapshot-id ...` | Show the statistics for the whole repo or the specified snapshots |
 | `resticw mount /mnt/restic`                       | Mount your remote repository                                      |
 
+## Setup macOS LaunchAgent
+LaunchAgent is the modern service scheduler in in macOS that uses [Launchd](https://www.launchd.info/).
+[Launchd](https://www.launchd.info/) is the modern built-in service scheduler in macOS. It has support for running services as root (Daemon) or as a normal user (Agent). Here we we set up an LauchAgent to be run as your normal user for starting regular backups.
+
+1. In general, follow the same setup as in (#setup-linux-systemd) except for:
+  * use `make install-launchagent` instead of `make install-systemd`
```
gerardbosch, line 268: _Disclaimer: I've not tried MacOS installation :)_

I was thinking that one of the things I like from root systemd installation is that configuration files, which contains credentials, are quite protected, as they are owned by root and not readable for others.

Any potential malware that runs under normal user permissions will never have access to that credentials, and that gives some sensation of security, specially if you take care about what do you run using `sudo`.

So, I'm not very sure if suggesting the use of `sudo` for MacOS could also bring additional security benefits, i.e. `sudo PREFIX=/usr/local make install-launchagent`. I have no idea if that would work with launchd/launchagent :)

erikw: I wanted to follow the typical Homebrew setup, which is without root. See https://docs.brew.sh/FAQ#why-does-homebrew-say-sudo-is-bad

Anyway, I'd say if someone has access to your system, that's your bigger problem. Gaining additional access to the backups is probably not going to be much worse than the hacker already having access to the real system. (Given that one uses unique B2 credentials  per bucket, which is recommended, so backups for other systems are not accessed via the token in `pw.txt`)

gerardbosch: Yep, fine! Just wanted to share my thoughts/curiosity, hehe. Is that I don't like much having readable credential files hanging around :sweat_smile: But I also agree the point of Homebrew about trust and sudo :smile:

PR #76, opened 2022-02-09
Add optional desktop notifications

gerardbosch:
When backing up a desktop system it is handy to have regular and
immediate feedback about backups. This features a notification of backup
stats summary, including the added size to the repository, so you can
have a quick follow-up about what are you uploading to the repository.

This may be very useful if you upload by accident a big file or directory
which should be in your exclusions, so it warns you and gives the chance
to fix the exclusions and remove the undesired snapshot later.

### DIFF
--- a/bin/restic_backup.sh
+++ b/bin/restic_backup.sh
@@ -66,18 +66,19 @@ done
 restic unlock &
 wait $!
 
-# Do the backup!
+# Do the backup! (and capture the output for further processing)
 # See restic-backup(1) or http://restic.readthedocs.io/en/latest/040_backup.html
 # --one-file-system makes sure we only backup exactly those mounted file systems specified in $RESTIC_BACKUP_PATHS, and thus not directories like /dev, /sys etc.
 # --tag lets us reference these backups later when doing restic-forget.
-{ backup_output=$(restic backup \
-	--verbose="$RESTIC_VERBOSITY_LEVEL" \
-	--one-file-system \
-	--tag "$RESTIC_BACKUP_TAG" \
-	--option b2.connections="$B2_CONNECTIONS" \
-	"${exclusion_args[@]}" \
-	"${extra_args[@]}" \
-	"${backup_paths[@]}" \
+{ backup_output=$( \
+	restic backup \
+		--verbose="$RESTIC_VERBOSITY_LEVEL" \
+		--one-file-system \
+		--tag "$RESTIC_BACKUP_TAG" \
+		--option b2.connections="$B2_CONNECTIONS" \
+		"${exclusion_args[@]}" \
+		"${extra_args[@]}" \
+		"${backup_paths[@]}" \
 	| tee /dev/fd/3 & )  # store output in var for further proc; also tee to a temp fd that's redirected to stdout
 } 3>&1
 wait $!
--- a/bin/restic_backup.sh
+++ b/bin/restic_backup.sh
@@ -66,21 +66,18 @@ done
 restic unlock &
 wait $!
 
-# Do the backup! (and capture the output for further processing)
+# Do the backup!
 # See restic-backup(1) or http://restic.readthedocs.io/en/latest/040_backup.html
 # --one-file-system makes sure we only backup exactly those mounted file systems specified in $RESTIC_BACKUP_PATHS, and thus not directories like /dev, /sys etc.
 # --tag lets us reference these backups later when doing restic-forget.
-{ backup_output=$( \
-	restic backup \
-		--verbose="$RESTIC_VERBOSITY_LEVEL" \
-		--one-file-system \
-		--tag "$RESTIC_BACKUP_TAG" \
-		--option b2.connections="$B2_CONNECTIONS" \
-		"${exclusion_args[@]}" \
-		"${extra_args[@]}" \
-		"${backup_paths[@]}" \
-	| tee /dev/fd/3 & )  # store output in var for further proc; also tee to a temp fd that's redirected to stdout
-} 3>&1
+restic backup \
+	--verbose="$RESTIC_VERBOSITY_LEVEL" \
+	--one-file-system \
+	--tag "$RESTIC_BACKUP_TAG" \
+	--option b2.connections="$B2_CONNECTIONS" \
+	"${exclusion_args[@]}" \
+	"${extra_args[@]}" \
+	"${backup_paths[@]}" &
 wait $!
 
 # Dereference and delete/prune old backups.
@@ -123,10 +120,18 @@ echo "Backup & cleaning is done."
 #
 if [ "$RESTIC_NOTIFY_BACKUP_STATS" = true ]; then
 	if [ -w "$RESTIC_BACKUP_NOTIFICATION_FILE" ]; then
-		added=$(echo "$backup_output" | grep -i 'Added to the repo:' | sed -E 's/.*dded to the repo: (.*)/\1/')
-		# sample:  processed N files, N.XYZ GiB in H:mm
-		size=$(echo "$backup_output" | grep  -i 'processed.*files,' | sed -E 's/.*rocessed.*files, (.*) in.*/\1/g')
-		echo "Added: ${added}. Snapshot size: ${size}" >> "$RESTIC_BACKUP_NOTIFICATION_FILE"
+		echo 'Notifications are enabled: Silently computing backup summary stats...'
+
+		snapshot_size=$(restic stats latest --tag "$RESTIC_BACKUP_TAG" | grep -i 'total size:' | cut -d ':' -f2 | xargs)  # xargs acts as trim
+		latest_snapshot_diff=$(restic snapshots --tag "$RESTIC_BACKUP_TAG" --latest 2 --compact \
+			| grep -Ei "^[abcdef0-9]{8} " \
+			| awk '{print $1}' \
+			| tr '\n' ' ' \
+			| xargs restic diff)
+        added=$(echo "$latest_snapshot_diff" | grep -i 'added:' | awk '{print $2 " " $3}')
+        removed=$(echo "$latest_snapshot_diff" | grep -i 'removed:' | awk '{print $2 " " $3}')
+
+		echo "Added: ${added}. Removed: ${removed}. Snap size: ${snapshot_size}" >> "$RESTIC_BACKUP_NOTIFICATION_FILE"
 	else
 		echo "[WARN] Couldn't write the backup summary stats. File not found or not writable: ${RESTIC_BACKUP_NOTIFICATION_FILE}"
 	fi
--- a/bin/restic_backup.sh
+++ b/bin/restic_backup.sh
@@ -70,14 +70,16 @@ wait $!
 # See restic-backup(1) or http://restic.readthedocs.io/en/latest/040_backup.html
 # --one-file-system makes sure we only backup exactly those mounted file systems specified in $RESTIC_BACKUP_PATHS, and thus not directories like /dev, /sys etc.
 # --tag lets us reference these backups later when doing restic-forget.
-restic backup \
+{ backup_output=$(restic backup \
 	--verbose="$RESTIC_VERBOSITY_LEVEL" \
 	--one-file-system \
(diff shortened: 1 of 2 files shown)


bin/restic_backup.sh
```diff
@@ -101,3 +103,31 @@ wait $!
...
+# TODO Clean/or rephrase the example comments below
+# In this case I'm running a user process that reads from a special pipe file and sends a desktop notification using
+# `notify-send`.
+# 
+# See: https://github.com/gerardbosch/dotfiles-linux/blob/main/home/.config/autostart/notification-queue.desktop and
+#      https://github.com/gerardbosch/dotfiles-linux/blob/main/home/bin/notification-queue-start-processing
+#
+if [ "$RESTIC_NOTIFY_BACKUP_STATS" = true ]; then
+	if [ -w "$RESTIC_BACKUP_NOTIFICATION_FILE" ]; then
+		added=$(grep -i 'Added to the repo:' <<< "$backup_output" | sed -E 's/.*dded to the repo: (.*)/\1/')
```
erikw, line 125: At least my editor (vim) becomes very confused with the here string syntax. I think it would be more readable as well using simple echo (as I think most people don't know abut thee here string feature).

```bash
added=$(echo "$backup_output" | grep -i 'Added to the repo:' | sed -E 's/.*dded to the repo: (.*)/\1/')
```

Is there some benefit of using the here string here? I assume the intention is to avoid "useless-echo", but it comes at a cost of readability

etc/restic/_global.env.sh
```diff
@@ -29,3 +29,8 @@ export RESTIC_BACKUP_EXTRA_ARGS=
 # Verbosity level from 0-3. 0 means no --verbose.
 # Override this value in a profile if needed.
 export RESTIC_VERBOSITY_LEVEL=0
+
+# (optional) Desktop notifications
```
erikw, line 33: Maybe add `. See backup.sh` for details on how to set this up.`

bin/restic_backup.sh
```diff
@@ -70,14 +70,16 @@ wait $!
 # See restic-backup(1) or http://restic.readthedocs.io/en/latest/040_backup.html
 # --one-file-system makes sure we only backup exactly those mounted file systems specified in $RESTIC_BACKUP_PATHS, and thus not directories like /dev, /sys etc.
 # --tag lets us reference these backups later when doing restic-forget.
-restic backup \
+{ backup_output=$(restic backup \
```
erikw, line 73: As I assume most people will maybe not use the notifications, this obscues the understandability of how this script works for new users that are just getting started with restic. 

Maybe we could instead extract this to a function `_caputre` so that we could keep this line simple. Here's a minimal example for how we can capture a `printf` to a variable `output`


```bash
_capture() {
  declare -n var="$1"
  shift
  local cmd=${@@Q}  # Quote $@ to preserve quoted strings. Ref. https://stackoverflow.com/a/39463371/265508
  var=$(eval $cmd)
}

_capture output \
	printf "test %s, %s\n" \
	"arg1" \
	"arg2"

echo "$output"
```

will print `test arg1, arg2`

erikw (changes requested):
I see the need for this! As it's optional. let's try to minimize the lines of code and attention it grabs from the reader though.

For macOS, I use [terminal-notifier](https://github.com/julienXX/terminal-notifier). Maybe we can also add in the comments that macOS users can make something analogous to your `notification-queue-start-processing`?

bin/restic_backup.sh
```diff
@@ -101,3 +103,31 @@ wait $!
...
+#
+# (optionally) Notify about backup summary stats.
+#
+# How to perform the notification is up to the user; the script only writes the info to the user-owned file in a fire
+# and forget fashion.
+#
+# One option to trigger desktop notifications on user-side is using a special FIFO file (a.k.a. pipe file), which will
+# work as a queue; plus, a user process to read from that queue and run a desktop notification command.
+#
+# TODO Clean/or rephrase the example comments below
```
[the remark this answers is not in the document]
gerardbosch: I left this in the draft PR with the TODO to clarify and show how it would work, hehe. I'll clean this in the next commit when I move the PR to final, now that the idea is clear :)

bin/restic_backup.sh
gerardbosch: Ups! I'm using VIM and didn't notice anything strange with the `<<<`, hehe. But I've updated it for the `echo ... | ` syntax :+1: :+1: 

I used the `<<<` as it looked shorter and probably saved one process (the `echo`), so made me quite sense. But after looking at it, I found that the "here string" is implemented with temporary files :-S https://stackoverflow.com/a/41899209/6108874

etc/restic/_global.env.sh
gerardbosch: Done!

bin/restic_backup.sh
gerardbosch: Oh! I see your point.

That's `{ output=$(command) | tee /dev/fd/3 & ) } 3>&1` the easiest way I found to do this.

I initially found another one that looked even more simple:
```bash
myvar=$( mycommand | tee /dev/tty | grep -c keyword )
```
ref: https://serverfault.com/a/989797/954276

But this failed when I run `sudo systemctl start restic-backup@default`, writing an error to the journal like `/dev/tty file not found` or something similar. Then I thought that tty is not present (I don't know how this works) when run by systemd.

---

Wow! I have to admit that I don't understand how the function you share `_capture` works under the hood :sweat_smile: it looks complex. But I totally agree that at the call-site it is very clean, and readable by using that function. But I'm suspecting we can have the same problem with `/dev/tty` when running it as a systemd service. Do you know how is that?

---

I see what you mean about UNIX philosophy, but thinking on extract the notification info from the journal, there are some points I'm not clear at all:

- How do we parse the right piece of the journal log? Should we use any date option to read the relevant piece of log? We will probably need to store the time at which the backup started.
- Where will we call the "notification/journal parsing" script? Do you mean that `backup.sh` will call it? That would be easy.
- BUT, relaying on `journalctl` leave the MacOS users out of notifications :( And the notification system is generic as only writes to a pipe/queue file.

Maybe just doing a little reformat on the current `{ output=$(command) | tee /dev/fd/3 & ) } 3>&1`, and adding line breaks to separate the `restic backup` from the rest of the construct, would make it more clear?

I'd like to listen what do you think is the better way. Oh! I'm not sure if most of the people is gonna use this to backup a server, but I guess every desktop user would like to have a notification about how many MB (or GB) did they back up since last day in case something got wrong :laughing:

gerardbosch: Hey! :) I swear that I had a similar PoC (more simplistic) weeks ago, based on `diff` as you show!! :D
But I finally discarded as it required an additional restic command (the `diff`), and I didn't know if it was worth it or coud incur in extra unnecessary charges or something similar. But I like that idea!

The only thing that's missing to me from the diff command is the snapshot size, which I like to have controlled. This can be easily achieved with `stats latest`, but involves an extra restic command.

I have no idea of issuing more restic commands comes at some non-negligible cost. If you think that's fine, maybe we can better use that approach. My original goal of reusing the `restic backup` output was to not run extra commands, just in case it was costly in any sense -which I'm not very sure.

Regarding the use of JSON, maybe is better to rely on plain output and parse with grep/sed to not add additional dependencies and increase the installation complexity (but I'm not oposed either, I already have `jq` :sweat_smile: ).

Let me know the overall and I can make adjustments with this :tada: :D

bin/restic_backup.sh
```diff
@@ -101,3 +104,30 @@ wait $!
...
+# In this case I'm running a user process that reads from a special pipe file and sends a desktop notification using
+# `notify-send`.
+#
+# See: https://github.com/gerardbosch/dotfiles-linux/blob/main/home/.config/autostart/notification-queue.desktop and
+#      https://github.com/gerardbosch/dotfiles-linux/blob/main/home/bin/notification-queue-start-processing
```
erikw, line 122: As I need to have this setup up on my machines as well, to be able to maintain everything that is integreated to this project, I created a setup for macOS that we can link to here

```
Examples of setup:
* Linux + notify-send
    * https://github.com/gerardbosch/dotfiles-linux/blob/main/home/bin/notification-queue-start-processing
    * https://github.com/gerardbosch/dotfiles-linux/blob/main/home/.config/autostart/notification-queue.desktop
* macOS + terminal-notifier
   * https://github.com/erikw/dotfiles/blob/personal/bin/notification-queue-notifier.sh
   * https://github.com/erikw/dotfiles/blob/personal/bin/com.user.notificationqueue.plist
```

erikw (changes requested):

bin/restic_backup.sh
```diff
@@ -101,3 +103,31 @@ wait $!
 #wait $!
 
 echo "Backup & cleaning is done."
+
+#
+# (optionally) Notify about backup summary stats.
```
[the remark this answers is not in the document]
gerardbosch: Done :)

bin/restic_backup.sh
gerardbosch: Check now :D
As a curiosity it takes 30s on my system to compute the backup stats and show the notification after the backup is finished.

bin/restic_backup.sh
```diff
@@ -101,3 +104,30 @@ wait $!
...
+#
+# How to perform the notification is up to the user; the script only writes the info to the user-owned file in a fire
+# and forget fashion.
+#
+# One option to trigger desktop notifications on user-side is using a special FIFO file (a.k.a. pipe file), which will
+# work as a queue; plus, a user process to read from that queue and run a desktop notification command.
+#
+# TODO Clean/or rephrase the example comments below
+# In this case I'm running a user process that reads from a special pipe file and sends a desktop notification using
```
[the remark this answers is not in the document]
gerardbosch: Sure! I wrote this mostly to you to review :) hence the TODO. I wanted early feedback about this feature before moving the PR from draft to ready.

bin/restic_backup.sh
gerardbosch: I added this, and changed the links with its permalink. This will reduce the chance of broken links, at the downside of not pointing the last version in case they change. But as they're placed here as example/doc I guess that having a link that "cannot" break is preferred.

gerardbosch:
Here is a sample of how notification look likes on Linux:

image

bin/restic_backup.sh
```diff
@@ -101,3 +101,37 @@ wait $!
...
+		snapshot_size=$(restic stats latest --tag "$RESTIC_BACKUP_TAG" | grep -i 'total size:' | cut -d ':' -f2 | xargs)  # xargs acts as trim
+		latest_snapshot_diff=$(restic snapshots --tag "$RESTIC_BACKUP_TAG" --latest 2 --compact \
+			| grep -Ei "^[abcdef0-9]{8} " \
+			| awk '{print $1}' \
+			| tr '\n' ' ' \
+			| xargs restic diff)
+        added=$(echo "$latest_snapshot_diff" | grep -i 'added:' | awk '{print $2 " " $3}')
+        removed=$(echo "$latest_snapshot_diff" | grep -i 'removed:' | awk '{print $2 " " $3}')
+
+		echo "Added: ${added}. Removed: ${removed}. Snap size: ${snapshot_size}" >> "$RESTIC_BACKUP_NOTIFICATION_FILE"
```
gerardbosch, line 133: I didn't add a longer body for the notification and condensed the message as much as I could on purpose. This is because at least on Linux, the notification message gets truncated if this is too long, here the "Snap size". I attached a screenshot in the conversation 😄

erikw (approved):

erikw:
Great, thank you!

gerardbosch:
You're welcome :)
Should we make a line on the README Optional section?

erikw:
2c14351cd4e7732bd7d8e2c0b8a48389e72ddee7

gerardbosch:
> Will do!
> 
> @gerardbosch could I include your screenshot above in the README?

Ups! I missed that message, hehe. Of course, you can use the screenshot for the README :D
I've seen you already put yours. Not sure if you want to show both Linux/Mac examples or just one es enough. Feel free to take it if you like :wink:

PR #84, opened 2022-02-16
Fix README/resticw commands

gerardbosch:
As per restic 0.12.1, the special snapshot ID `latest` is not supported by `diff` command.


README.md
```diff
@@ -517,8 +517,8 @@ Useful commands:
 | Command                                           | Description                                                       |
 |---------------------------------------------------|-------------------------------------------------------------------|
 | `resticw snapshots`                               | List backup snapshots                                             |
-| `resticw diff <snapshot-id> latest`               | Show the changes from the latest backup                           |
```
gerardbosch, line 520: `latest` is not supported here :(

PR #85, opened 2022-02-16
Add docopt.sh option: DOCOPT_OPTIONS_FIRST=true

gerardbosch:
This will allow to use `resticw` as a true wrapper, enabling to pass
restic options through it, for example:

```
resticw --profile profileX snapshots --compact
```

will now be suported by the wraper :)

Relates to https://github.com/erikw/restic-automatic-backup-scheduler/pull/71


erikw:
Great! I tried it out -- works :)

PR #86, opened 2022-02-16
Redirect stderr to systemd journal

gerardbosch:
Errors were not written to the journal as systemd-cat was only reading
stdin. Now, errors printed by restic are shown in the journal.

### DIFF
--- a/usr/lib/systemd/system/restic-backup@.service
+++ b/usr/lib/systemd/system/restic-backup@.service
@@ -12,4 +12,4 @@ Nice=10
 Environment="HOME=/root"
 # The random sleep (in seconds) is in the case of multiple backup profiles. Many restic instances started at the same time could case high load or network bandwith usage.
 # `systemd-cat` allows showing the restic output to the systemd journal
-ExecStart=bash -c 'ps cax | grep -q restic && sleep $(shuf -i 0-300 -n 1); source $INSTALL_PREFIX/etc/restic/%I.env.sh && $INSTALL_PREFIX/bin/restic_backup.sh | systemd-cat'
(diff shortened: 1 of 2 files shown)


DianQK:
Looks like this PR breaks the Email Notification on Failure feature?

```
$ some-bad-script | systemd-cat
bash: some-bad-script: command not found...
$ echo $?
0
```

It looks like this will make the backup script always output a successful result?
Please let me know if I'm wrong.

gerardbosch:
Hi there! 
I'm answering from my phone now but I guess you're right. Although I'm not familiar on the email notifications right now.

Maybe we need to do:
set -o pipefail

on that systemd script to make it fail? Let's see what @erikw says :)

Cheers!

erikw:
Hmm did not think of that before


@DianQK could you try changing
`usr/lib/systemd/system/restic-backup@.service` like this:

```diff
- ExecStart=bash -c 'ps cax | grep -q restic && sleep $(shuf -i 0-300 -n 1); source $INSTALL_PREFIX/etc/restic/%I.env.sh && $INSTALL_PREFIX/bin/restic_backup.sh 2>&1 | systemd-cat
+ ExecStart=bash -c 'set -o pipefail; ps cax | grep -q restic && sleep $(shuf -i 0-300 -n 1); source $INSTALL_PREFIX/etc/restic/%I.env.sh && $INSTALL_PREFIX/bin/restic_backup.sh 2>&1 | systemd-cat
```

and see if that works with the `OnFailure=status-email-user@%n.service` ?

DianQK:
@gerardbosch @erikw 
It works.
btw, I used gotify instead of email.

PR #87, opened 2022-02-16
Update README: Metered connection check

gerardbosch:
Add a single command to copy and paste to install the metered connection
check.


erikw:
Thank you. I did a slight reformat to be consistent with the style of the other sections, but kept the 1-copy-paste command  :)

PR #107, opened 2023-03-05
pre_backup commands

antedebaas:
i needed a feature to preform command prior to actually backing up.
in my case some mysqldumps.
i thought this could help others too.

### DIFF
--- a/bin/restic_backup.sh
+++ b/bin/restic_backup.sh
@@ -46,7 +46,7 @@ IFS=' ' read -ra extra_args <<< "$RESTIC_BACKUP_EXTRA_ARGS"
 
 #run the commands listed here before running the actual backup.
 PRE_SCRIPT="${INSTALL_PREFIX}/etc/restic/pre_backup.sh"
-test -x "$PRE_SCRIPT" && $PRE_SCRIPT
+test -x "$PRE_SCRIPT" && "$PRE_SCRIPT"
 
 # Set up exclude files: global + path-specific ones
 # NOTE that restic will fail the backup if not all listed --exclude-files exist. Thus we should only list them if they are really all available.
--- a/bin/restic_backup.sh
+++ b/bin/restic_backup.sh
@@ -47,7 +47,7 @@ IFS=' ' read -ra extra_args <<< "$RESTIC_BACKUP_EXTRA_ARGS"
 #run the commands listed here before running the actual backup.
 while IFS= read -r command; do
     $command
-done < /etc/restic/pre_backup.txt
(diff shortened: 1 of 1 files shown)


bin/restic_backup.sh
```diff
@@ -44,6 +44,10 @@ assert_envvars \
 IFS=':' read -ra backup_paths <<< "$RESTIC_BACKUP_PATHS"
 IFS=' ' read -ra extra_args <<< "$RESTIC_BACKUP_EXTRA_ARGS"
 
+#run the commands listed here before running the actual backup.
+while IFS= read -r command; do
```
erikw, line 48: Hm what are the benefits of reading the commands as a text file with splitting, instead of having a shell script that is sourced or executed directly?

```bash
PRE_SCRIPT="$INSTALL_PREFIX/etc/restic/pre_backup.sh"
test -x "$PRE_SCRIPT" && $PRE_SCRIPT
# -or-
test -e "$PRE_SCRIPT" && . $PRE_SCRIPT
```

antedebaas: actually none. :)

bin/restic_backup.sh
```diff
@@ -44,6 +44,9 @@ assert_envvars \
 IFS=':' read -ra backup_paths <<< "$RESTIC_BACKUP_PATHS"
 IFS=' ' read -ra extra_args <<< "$RESTIC_BACKUP_EXTRA_ARGS"
 
+#run the commands listed here before running the actual backup.
```
erikw, line 47:
```suggestion
# If you need to run some commands before performing the backup; create this file, put them there and make the file executable.
```

bin/restic_backup.sh
```diff
@@ -44,6 +44,9 @@ assert_envvars \
...
+PRE_SCRIPT="${INSTALL_PREFIX}/etc/restic/pre_backup.sh"
+test -x "$PRE_SCRIPT" && $PRE_SCRIPT
```
erikw, line 49:
```suggestion
test -x "$PRE_SCRIPT" && "$PRE_SCRIPT"
```

erikw (changes requested):
Let's remove the checked in `etc/restic/pre_backup.txt` as the instructions explains that one should create it when needed :)

antedebaas:
agreed :)

antedebaas:
can you give me a tag on this?
that way my ansible script can use your branch again

erikw:
@antedebaas here you go :)

https://github.com/erikw/restic-automatic-backup-scheduler/releases/tag/v7.4.0

PR #112, opened 2023-05-21
Allow spaces in extra args

### LINKED ISSUE #111: Allow space in extra args
Hi!

It seems, `IFS=' ' read -ra extra_args <<< "$RESTIC_BACKUP_EXTRA_ARGS"` does not respect escaped spaces.

If you print out the entries in `"${extra_args[@]}"` you can see that they are split:
```
--option
sftp.command='ssh\
-F\
/etc/restic/ssh_config\
restic-backup-host\
-s\
sftp'
--exclude-caches
--one-file-system
```
In this example, it leads to -F (option for sftp) to be used as option for restic.

Not sure how to resolve - use `RESTIC_BACKUP_EXTRA_ARGS` somehow like an array? Or allow escapes?

erikw:
Fixes #111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Warn on certain unset envvars instead of error-exit.
+- Allow escaped spaces in EXTRA_ARGS.
 
 ## [7.4.0] - 2023-03-08
 ### Added
--- a/bin/restic_backup.sh
+++ b/bin/restic_backup.sh
@@ -9,7 +9,7 @@
 #   $ restic_backup.sh
 
 # Exit on error, unset var, pipe failure
-set -euo pipefail
+set -eo pipefail
 
 # Clean up lock if we are killed.
 # If killed by systemd, like $(systemctl stop restic), then it kills the whole cgroup and all it's subprocesses.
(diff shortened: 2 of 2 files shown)


nettnikl:
I don't have permission to push here.. But i think i might found a way: https://github.com/erikw/restic-automatic-backup-scheduler/pull/113

nettnikl:
Btw, we have to filter out some args... Some args do not work for unlock, like `unknown flag: --exclude-caches`

erikw:
> Btw, we have to filter out some args... Some args do not work for unlock, like `unknown flag: --exclude-caches`

Why does the unlock command need extra args?

erikw:
> I don't have permission to push here.. But i think i might found a way: #113

Oh forgot to clean up the testing, thus the linter job complained as well  :P

The PR is updated, please try it again :)

nettnikl:
> I'd say this is such a special case that it's bloating the code too much, to solve one persons specific setup. See https://github.com/erikw/restic-automatic-backup-scheduler#project-scope I'd recommend you to keep a personal fork that adds extra args to the unlock command as well for your specific setup :)

Okay, will keep my fork updated then :+1: 

Had trouble with the given changes still, though: `unknown flag: --one-file-system --exclude-caches --option sftp.command`
Still same issue as described here: https://github.com/erikw/restic-automatic-backup-scheduler/issues/111#issuecomment-1556167426

erikw:
> Had trouble with the given changes still, though: `unknown flag: --one-file-system --exclude-caches --option sftp.command`
> Still same issue as described here: [#111 (comment)](https://github.com/erikw/restic-automatic-backup-scheduler/issues/111#issuecomment-1556167426)

Ha, interesting! With my debug printing, it looks like your command was showing properly and would work. I tried to stay by the simpler solution. 

I applied to solution you demonstrated in #113 to this PR now.

Does it work now? :)

nettnikl:
Yeah, should be fine, it looks to be similar to my earlier proposals (minus my `restic unlock` mod, plus docu), so i'm happy with it!

erikw:
> it looks to be similar to my earlier proposals

Indeed, I copied it! :)

PR #114, opened 2023-05-24
Fix invalid syntax

nettnikl:
Fix `invalid argument "" for "-H, --keep-hourly" flag: strconv.ParseInt: parsing "": invalid syntax`


erikw:
Oh that was missed to check, thx!

PR #115, opened 2023-05-24
Allow extra args for non-backup commands

nettnikl:
Add the env variable `RESTIC_EXTRA_ARGS` that will be added to _all_ `restic` commands, including `restic stats` etc. This can be used for additional setting for accessing the repo (like the sftp binary setting).

### DIFF
--- a/bin/restic_backup.sh
+++ b/bin/restic_backup.sh
@@ -8,8 +8,9 @@
 #   $ source $PREFIX/etc/default.env.sh
 #   $ restic_backup.sh
 
-# Exit on error, unset var, pipe failure
-set -eo pipefail
+set -o errexit
+set -o pipefail
+[[ "${TRACE-0}" =~ ^1|t|y|true|yes$ ]] && set -o xtrace
 
 # Clean up lock if we are killed.
 # If killed by systemd, like $(systemctl stop restic), then it kills the whole cgroup and all it's subprocesses.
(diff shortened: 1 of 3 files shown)


nettnikl:
You mentioned it's probably too specific, but i'd like to drop this mr anyways, it's quite lightweight.

bin/restic_backup.sh
```diff
@@ -108,7 +114,7 @@ restic backup \
 	--tag "$RESTIC_BACKUP_TAG" \
 	"${B2_ARG[@]}" \
 	"${exclusion_args[@]}" \
-	"${extra_args[@]}" \
```
erikw, line 111: As my understanding is that `RESTIC_EXTRA_ARGS` should be applied to all restic commands, and `RESTIC_BACKUP_EXTRA_ARGS` only to the backup command, we should keep `"${extra_args[@]}" ` here

erikw (commented):
If I get the intention right, `RESTIC_EXTRA_ARGS` would be used to set "Global Flags" as described at

https://restic.readthedocs.io/en/latest/manual_rest.html

Is this correct?

I think we should add commented-out example to 
https://github.com/erikw/restic-automatic-backup-scheduler/blob/main/etc/restic/default.env.sh
just as there is for `RESTIC_BACKUP_EXTRA_ARGS` already.

nettnikl:
Hey @erikw , sorry for the delay, i had a stressful time at work. Mind to take a look at the updates?

erikw:
No worries, I understand :).

I fixed indentation and Changelog just in the PR. Thanks!

PR #118, opened 2023-08-02
Update README.md: add $RESTIC_EXTRA_ARGS

svengo:
Add `$RESTIC_EXTRA_ARGS` to all restic calls.


erikw:
RESTIC_EXTRA_ARGS doesn't exist for a new user, thus makes no sense e.g. when $ restic init. 

The profile must be sourced before this env var is available as well. To keep it simple, maybe we could add a notice that if one uses $RESTIC_EXTRA_ARGS, then one can include these in the command lines. I think few people depend on this , so I rather keep it simple and not daunting for the general user:)

PR #121, opened 2023-09-10
Do not escape parsed args, fixes #120

### LINKED ISSUE #120: [BUG] Retention periods are being ignored by restic forget when "${extra_args[@]}" is empty
When the variable  "${extra_args[@]}" is empty, the retention period options will be ignored and this warning will show up 

```shell
Ignoring "": multiple IDs with prefix "" found
Ignoring "filters": explicit snapshot ids are given
``` 

### I was able to solve my issue by removing the "${extra_args[@]}"  variable from restic forget in restic_backup.sh

```python
restic forget \
        --verbose="$RESTIC_VERBOSITY_LEVEL" \
        --tag "$RESTIC_BACKUP_TAG" \
        "${B2_ARG[@]}" \
        "${extra_args[@]}" \       <======== REMOVED THIS to solve my issue
        --prune \
        --group-by "paths,tags" \
        --keep-hourly "$RESTIC_RETENTION_HOURS" \
        --keep-daily "$RESTIC_RETENTION_DAYS" \
        --keep-weekly "$RESTIC_RETENTION_WEEKS" \
        --keep-monthly "$RESTIC_RETENTION_MONTHS" \
        --keep-yearly "$RESTIC_RETENTION_YEARS" &
wait $!
``` 

- OS: Ubuntu 22.04.2 LTS
- Restic Version: 0.16.0 compiled with go1.20.6 on linux/amd64

nettnikl:


nettnikl:
Should we do the same in exclusion_args, backup_extra_args, backup_paths?

erikw:
@ggzica could you confirm that this works for you as well? 

You can check out this PR in your local git clone with the GitHub `gh` command: `$ gh pr checkout 121`

ggzica:
Hi,
I confirm that this works for me as well.

image

nettnikl:
One thing also, will this break compat with spaces in args? will we ever have spaces in args?

erikw:
> One thing also, will this break compat with spaces in args? will we ever have spaces in args? In those cases, i might just reescape the params in this array, so when expanding them without quotes, they result in correct params.


Yes there will definitely be space in the args, it's even in the example

https://github.com/erikw/restic-automatic-backup-scheduler/blob/3f05091edbc4e186b94135f5a5e7df31af5e0771/etc/restic/default.env.sh#L49

So this must work!

> How about switching to python?
Out of scope ;)

PR #129, opened 2023-12-12
feat(resticw): Add a switch to run a diff on the latest snapshot

gerardbosch:
There's no obvious way to simply run a restic diff for the latest snapshot against the previous one.

This simplifies the task to inspect the last snapshot changes. The user only needs to run `sudo resticw --diff-latest` (optionally specifying a profile).

### DIFF
--- a/bin/resticw
+++ b/bin/resticw
@@ -1,19 +1,24 @@
 #!/usr/bin/env bash
 # @generated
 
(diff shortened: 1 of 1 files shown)


gerardbosch:
Oops! I've just seen the announcement 😅😅, would you still accept this last minute small commit?

erikw (approved):
Yes of course — you had no forewarning =)

PR #131, opened 2023-12-12
fix: Change INSTALL_PREFIX placeholder notation to templating notation

gerardbosch:
The former shell notation `$INSTALL_PREFIX` could result confusing. As
this is a placeholder that's replaced by `make`, we switch it to a
template notation `{{ INSTALL_PREFIX }}`.

Also, fixes one occurrence of the placeholder in `restic_backup.sh` not replaced (was using `${}` notation and not replaced by `sed`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,14 @@
 #### Notes ####################################################################
 # This build process is done in three stages (out-of-source build):
 # 1. copy source files to the local build directory.
-# 2. build dir: replace the string "$INSTALL_PREFIX" with the value of $PREFIX
+# 2. build dir: replace the string "{{ INSTALL_PREFIX }}" with the value of $PREFIX
 # 3. install files from the build directory to the target directory.
 #
 # Why this dance?
 # * To fully support that a user can install this project to a custom path e.g.
 #   $(PREFIX=/usr/local make install), we need to modify the files that refer
 #   to other files on disk. We do this by having a placeholder
-#   "$INSTALL_PREFIX"  that is substituted with the value of $PREFIX when
+#   "{{ INSTALL_PREFIX }}"  that is substituted with the value of $PREFIX when
 #   installed.
 # * We don't want to modify the files that are controlled by git, thus let's
 #   copy them to a build directory and then modify.
@@ -57,7 +57,7 @@ LAUNCHAGENT_CHECK			= com.github.erikw.restic-check
 LAUNCHAGENT_TARGET_BACKUP	= gui/$(UID)/$(LAUNCHAGENT_BACKUP)
 LAUNCHAGENT_TARGET_CHECK	= gui/$(UID)/$(LAUNCHAGENT_CHECK)
 
-# What to substitute $INSTALL_PREFIX in sources to.
+# What to substitute {{ INSTALL_PREFIX }} in sources to.
 # This can be useful to set to empty on commandline when building e.g. an AUR
 # package in a separate build directory (PREFIX).
 INSTALL_PREFIX := $(PREFIX)
@@ -192,11 +192,11 @@ install-targets-schedtask: $(BUILD_DIR_SCHEDTASK)/$(SCHEDTASK_INSTALL)
 uninstall-targets-schedtask: $(BUILD_DIR_SCHEDTASK)/$(SCHEDTASK_UNINSTALL)
 	test $(CUR_OS) != Windows || ./$<
 
-# Copies sources to build directory & replace "$INSTALL_PREFIX".
+# Copies sources to build directory & replace "{{ INSTALL_PREFIX }}".
 $(BUILD_DIR)/% : %
 	@${MKDIR_PARENTS} $@
 	cp $< $@
-	sed -i.bak -e 's|$$INSTALL_PREFIX|$(INSTALL_PREFIX)|g' $@; rm $@.bak
+	sed -i.bak -e 's|{{ INSTALL_PREFIX }}|$(INSTALL_PREFIX)|g' $@; rm $@.bak
 
 # Install destination script files.
 $(DEST_DIR_SCRIPT)/%: $(BUILD_DIR_SCRIPT)/%
--- a/README.md
+++ b/README.md
@@ -308,7 +308,7 @@ $ git clone https://github.com/erikw/restic-automatic-backup-scheduler.git && cd
 
 Make a quick search-and-replace in the source files:
 ```console
-$ find etc bin -type f -exec sed -i.bak -e 's|{{ INSTALL_PREFIX }}||g' {} \; -exec rm {}.bak \;
+$ find . -type f -exec sed -i.bak -e 's|{{ INSTALL_PREFIX }}||g' {} \; -exec rm {}.bak \;
 ```
 and you should now see that all files have been changed like e.g.
 ```diff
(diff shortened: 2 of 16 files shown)


gerardbosch:
Most of the changes are a massive replace of `$INSTALL_PREFIX` to `{{ INSTALL_PREFIX }}` as commented here https://github.com/erikw/restic-automatic-backup-scheduler/pull/127#issuecomment-1851010961, but I split the PR in atomic commits to ease the review.

Makefile
```diff
@@ -192,11 +192,11 @@ install-targets-schedtask: $(BUILD_DIR_SCHEDTASK)/$(SCHEDTASK_INSTALL)
 uninstall-targets-schedtask: $(BUILD_DIR_SCHEDTASK)/$(SCHEDTASK_UNINSTALL)
 	test $(CUR_OS) != Windows || ./$<
 
-# Copies sources to build directory & replace "$INSTALL_PREFIX".
+# Copies sources to build directory & replace "{{ INSTALL_PREFIX }}".
 $(BUILD_DIR)/% : %
 	@${MKDIR_PARENTS} $@
 	cp $< $@
-	sed -i.bak -e 's|$$INSTALL_PREFIX|$(INSTALL_PREFIX)|g' $@; rm $@.bak
+	sed -i.bak -e 's|{{ INSTALL_PREFIX }}|$(INSTALL_PREFIX)|g' $@; rm $@.bak
```
gerardbosch, line 199: This was the important piece of the replace, removing the extra `$` on the sed expression as was previously necessary to escape the dollar.

bin/restic_backup.sh
```diff
@@ -97,7 +97,7 @@ B2_ARG=
 [ -z "${B2_CONNECTIONS+x}" ] || B2_ARG=(--option b2.connections="$B2_CONNECTIONS")
 
 # If you need to run some commands before performing the backup; create this file, put them there and make the file executable.
-PRE_SCRIPT="${INSTALL_PREFIX}/etc/restic/pre_backup.sh"
```
gerardbosch, line 100: Here was the issue with the prefix, notice previous `${}` notation.

README.md
```diff
@@ -484,7 +484,7 @@ Put this file in `/bin`:
 * `systemd-email`: Sends email using sendmail(1). This script also features time-out for not spamming Gmail servers and getting my account blocked.
 
 Put this file in `/etc/systemd/system/`:
-* `status-email-user@.service`: A service that can notify you via email when a systemd service fails. Edit the target email address in this file.
+* `status-email-user@.service`: A service that can notify you via email when a systemd service fails. Edit the target email address in this file, and replace or remove `{{ INSTALL_PREFIX }}` according to your installation.
```
erikw, line 487: Installing with the Makefole should not require to manually edit this file, right?

README.md
```diff
@@ -308,17 +308,17 @@ $ git clone https://github.com/erikw/restic-automatic-backup-scheduler.git && cd
 
 Make a quick search-and-replace in the source files:
 ```console
-$ find etc bin -type f -exec sed -i.bak -e 's|$INSTALL_PREFIX||g' {} \; -exec rm {}.bak \;
+$ find . -type f -exec sed -i.bak -e 's|{{ INSTALL_PREFIX }}||g' {} \; -exec rm {}.bak \;
```
erikw, line 311: Maybe don’t like https://github.com/erikw/restic-automatic-backup-scheduler/pull/127/files and add ‘use’ just, as doing ‘.’ Would also modify README.md I would be.

README.md
gerardbosch: Nope, as @giuaig pointed out in the other talk, this is an optional feature and intended to be manually installed according to the README. It is totally ignored by `make` 😇 right now.

I'll address your other comments later!

erikw: Ah gotcha!

README.md
gerardbosch: Ooops, you're right! I forgot about the README and even Makefile 🤦.

This is the list of files where the placeholder is found, there's also `Library` and `ScheduledTask` in there:

```console
❯ rg '\{\{ INSTALL_PREFIX \}\}' --files-with-matches --hidden | sort
bin/cron_mail
bin/restic_backup.sh
bin/restic_check.sh
bin/resticw
etc/cron.d/restic
etc/restic/default.env.sh
etc/restic/_global.env.sh
.git/COMMIT_EDITMSG
.git/logs/HEAD
Library/LaunchAgents/com.github.erikw.restic-backup.plist
Library/LaunchAgents/com.github.erikw.restic-check.plist
Makefile
README.md
ScheduledTask/install.ps1
usr/lib/systemd/system/nm-unmetered-connection.service
usr/lib/systemd/system/restic-backup@.service
usr/lib/systemd/system/restic-check@.service
usr/lib/systemd/system/status-email-user@.service
```

I post a couple of alternatives I see here (below), you can accept here the one you find more suited if any :) After looking around it I think option (A) may be ok. Notice that this is only documentation for the manual setup, hehe.

**Option A. Explicit exclusion (using grep to find matches)**
(Based on the list above) Exclude hidden dirs, readme and makefile from the search.
```suggestion
$ grep -r -l --exclude-dir '.*' --exclude 'Makefile' --exclude 'README.md' --exclude 'CHANGELOG.md' '{{ INSTALL_PREFIX }}' | xargs sed -i -e 's|{{ INSTALL_PREFIX }}||g'
```

This is the output of the first part (before going into `sed -i`):

```console
❯ grep -r -l --exclude-dir '.*' --exclude 'Makefile' --exclude 'README.md' --exclude 'CHANGELOG.md' '{{ INSTALL_PREFIX }}'                                              
bin/resticw
bin/restic_backup.sh
bin/cron_mail
bin/restic_check.sh
ScheduledTask/install.ps1
Library/LaunchAgents/com.github.erikw.restic-check.plist
Library/LaunchAgents/com.github.erikw.restic-backup.plist
usr/lib/systemd/system/status-email-user@.service
usr/lib/systemd/system/restic-backup@.service
usr/lib/systemd/system/restic-check@.service
usr/lib/systemd/system/nm-unmetered-connection.service
etc/restic/default.env.sh
etc/restic/_global.env.sh
etc/cron.d/restic
```

**Option B. Explicit inclusion**
(I'm understanding that `Library` and `ScheduledTask` need to take part of this)
```suggestion
$ find bin etc usr Library ScheduledTask -type f -exec sed -i.bak -e 's|{{ INSTALL_PREFIX }}||g' {} \; -exec rm {}.bak \;
```

But this would leave `sed` to deal with non-text files in other directories created in the future like images added in the future.

---

I guess a different option would be to create a directory `src/` and put all files in there so they get isolated from README, Makefile, etc —but that would be out of the scope of the PR, hehe.

gerardbosch: I added co-author info in one commit and force-pushed, but if you commit straight from this button either option-a or b, you can add it too:

```
co-authored-by: giuaig <13609224+giuaig@users.noreply.github.com>
```

erikw: Thanks for investigating!

Option A is fine, if it works also with BSD Gero that is default in macOS. I’m unable to verify this at the moment (you know why), but if you can verify it works on Mac too , then let’s go :)

gerardbosch: I could test it and ups! It seems that there were inconsistencies in both `grep` and `sed` 😱, really unfortunate! The grep had solution, but the sed I'm afraid don't have or couldn't come up with it.

Let's go for option-b then?

erikw: Thank you for checking. Option B applied!

PR #132, opened 2023-12-13
FIX Readme typo on 'Email Notification'

giuaig:
The services are `restic-backup` and `restic-check`.
Parameter OnFailure is not present in `status-email-user@.service` ;)


erikw:
Thanks!

PR #144, opened 2025-08-08
Check that there are enough snapshots before diff

johnmaguire:
The first run after init'ing tends to cause a systemd failure due to not having enough snapshots for statistics.

### DIFF
--- a/bin/restic_backup.sh
+++ b/bin/restic_backup.sh
@@ -170,6 +170,19 @@ if [[ -n "$RESTIC_BACKUP_STATS_DIR" || -n "$RESTIC_BACKUP_NOTIFICATION_FILE" ]];
 		| awk '{print $1}' \
 		| tail -2 \
 		| tr '\n' ' ')
+
+	snapshot_count=$(echo "$latest_snapshots" | wc -l)
+	if [[ $snapshot_count -lt 2 ]]; then
+		echo "Warning: $snapshot_count snapshot(s) found. Skipping diff stats (need at least 2 snapshots)."
+	else
+		latest_snapshot_diff=$(echo "$latest_snapshots" | xargs restic diff)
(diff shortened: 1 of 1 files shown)


bin/restic_backup.sh
```diff
@@ -170,6 +170,19 @@ if [[ -n "$RESTIC_BACKUP_STATS_DIR" || -n "$RESTIC_BACKUP_NOTIFICATION_FILE" ]];
...
+	else
+		latest_snapshot_diff=$(echo "$latest_snapshots" | xargs restic diff)
+		added=$(echo "$latest_snapshot_diff" | grep -i 'added:' | awk '{print $2 " " $3}')
+		removed=$(echo "$latest_snapshot_diff" | grep -i 'removed:' | awk '{print $2 " " $3}')
+		snapshot_size=$(restic stats latest --tag "$RESTIC_BACKUP_TAG" | grep -i 'total size:' | cut -d ':' -f2 | xargs)  # xargs acts as trim
+		snapshotId=$(echo "$latest_snapshots" | cut -d ' ' -f2)
+		statsMsg="Added: ${added}. Removed: ${removed}. Snap size: ${snapshot_size}"
+	fi
+
 	latest_snapshot_diff=$(echo "$latest_snapshots"	| xargs restic diff)
```
erikw, line 186: It looks like you forgot to remove the snapshot calculation lines, that are now copied to the `else` branch.


Probably the `echo "$statsMsg"` and 2 lines below should also be in the new `else`?

erikw (changes requested):
Good idea! Just some git diff questions to review.

bin/restic_backup.sh
johnmaguire: Hey, sorry about that! I have a slightly modified copy of this in my Ansible repository and I forgot to remove the snapshot calculation lines when moving it over to this repository.

Good point on the following two lines - when I tested it as-is, no errors were returned, but it doesn't make sense to call these functions without the data.

